PR #43, opened 2015-04-14
comment out unused variable.

lolisa:


ldionne:
First, thanks for contributing! I've been using the `-Wno-unused-parameter` flag since the beginning of the project, and I did not realize it would show up as warnings in user code. That's a good catch. Your PR also made me realize I had a bug in the constexpr algorithms, which is now fixed by 06126249.

I made some tweaks:
- Removed the `-Wno-unused-parameter` flag from CMake
- Removed more unused parameter warnings from the tests
- Rebased on top of develop

The PR is currently being tested on Travis (846cfe9) to make sure there are no more unused parameter warnings, and once that's done I'll merge.

Thanks!

PR #168, opened 2015-09-01
[Build] Check C++ ABI and exception issues with Clang/GCC on Cygwin

### LINKED ISSUE #167: Boost::Hana compiles on Cygwin with Clang 3.5.2 with some changes
Hi Louis,
after some tweaking, Boost::Hana successfully compiles most of the tests and examples on Cygwin.
There were two major issues: 
1) Clang 3.5.2 currently is not able to compile with exceptions, thus requires the flap -fno-exceptions. Accordingly, all tests and examples using exceptions fail.
2) Clang also builds with the C++ standard library provided by g++ 4.9.3. Hence, the type traits std::is_trivially_*, the struct aligned_union_t, and some examples using std::to_string fail.

Unfortunately ruby 1.8 is too old in the default cygwin package, so benchmarks are not working.
Other than that, everything seems fine. If you want, I can provide some patches preferably via email; not yet that much experience with github pull requests...

Cheers, Markus.

Cygwin 64 on Windows 7
clang version 3.5.2 (tags/RELEASE_352/final)
Target: x86_64-unknown-windows-cygnus, Thread model: posix
g++ (GCC) 4.9.3 providing libstdc++6 (version 4.9.3-1) and supposedly used for linking
cmake version 3.3.1

m-j-w:
Fixes  #167

- Check whether compiler supports and successfully compiles with exceptions.
- Check whether the provided stdlib has C++14 type traits fully implemented.
- Perform some compiler version checks in cmake and give hints to the user.

Both errors occured on Cygwin using Clang 3.5.2 with libstdc++ from GCC
4.9.3. However, Clang may also use the GCC stdlib on other systems, like
Linux Mint.
The error with exceptions is also expected to occure on MSYS2 and maybe
Windows, since Clang seems to have general problems there.

Unfortunately, the incomplete stdlib is supposedly not detectable from within
the compilation. Thus a test is created which may fail during compilation.
If so, an additional symbol is defined to handle the issue:
BOOST_HANA_CXX_INCOMPLETE_STDLIB

Availability of exceptions is tested with the existance of __EXCEPTIONS, and
handled accordingly. Instead of throw, a assert(false) or abort() comes to mind.

Lastly, the cmake cygwin warning message is deactivated in CMakeLists.txt,
and the main include <boost/hana/hana.hpp> checks the C++ standard to be
at least C++14 or raises an error otherwise. Should give beginners a valuable
hint on what is going wrong...

A side note for other Cygwin/MinGW/MSYS2 users: Compile with Clang or
GCC with static binding of (both) standard libs using 
     -static-libgcc -static-libstdc++
Makes your life a lot easier.

Remaining issues are some non-compiling tests and examples. These have been
removed from the build on Cygwin. The reason may also be the incomplete stlib.
See test/CMakeLists.txt and example/CMakeLists.txt

Tested on
- Cygwin/Clang 3.5.2/libstdc++ of gcc 4.9.3 w/o exceptions (success)
- Cygwin/Clang 3.5.2/libstdc++ of gcc 4.9.3 w/ exceptions (failure, but captured)
- Cygwin/GCC 4.9.3/libstdc++ of gcc 4.9.3 (failure, as expected)
- Linux Mint/Clang 3.6.0/libstdc++ of gcc 4.8.4 (failure, as expected)
- Linux Mint/Clang 3.6.0/libc++ of clang 3.6.0 w/ exceptions (success)
- Linux Mint/gcc 4.8.4/libstdc++ of gcc 4.8.4 (uncaptured failure, as expected)

### DIFF
--- a/cmake/CheckCxxCompilerCompatibility.cmake
+++ b/cmake/CheckCxxCompilerCompatibility.cmake
@@ -1,4 +1,5 @@
 # Copyright Louis Dionne 2015
+# with contributions of Markus J. Weber 2015
 # Distributed under the Boost Software License, Version 1.0.
 # (See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
 #
@@ -61,4 +62,3 @@ if(SHOW_PREFERRED_COMPILER)
       ###
     ")
 endif()
-
--- a/cmake/CheckCxxCompilerCompatibility.cmake
+++ b/cmake/CheckCxxCompilerCompatibility.cmake
@@ -1,0 +1,64 @@
+# Copyright Louis Dionne 2015
+# Distributed under the Boost Software License, Version 1.0.
+# (See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
+#
+#
+# This CMake module checks the compiler and linker for compatibility and
+# provides hints to the user.
+
+if    (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang")
+   if (${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS "3.5.0")
+      message(WARNING "
+      ###
+      ###   You appear to use Clang++ compiler version ${CMAKE_CXX_COMPILER_VERSION},
+      ###   which is most likely unable to compile Boost::Hana.
+      ###
+      ")
+      SET(SHOW_PREFERRED_COMPILER ON)
+    endif()
+elseif(${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU")
+    SET(SHOW_PREFERRED_COMPILER ON)
+    if (${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS "5.1.0")
+      message("
+      ###
+      ###   You appear to use GNU C++ compiler version ${CMAKE_CXX_COMPILER_VERSION},
+      ###   which is known to be unable to compile Boost::Hana.
+      ###
+      ###   You should at least upgrade to version >= 5.1.0
+      ###
+      ")
+    else()
+      message("
+      ###
+      ###   You appear to use GNU C++ compiler version ${CMAKE_CXX_COMPILER_VERSION}.
+      ###
+      ###   The compatibility with Boost::Hana is not yet confirmed.
+      ###
+      ")
+    endif()
+else()
+      message("
+      ###
+      ###   NOTE: You appear to use a compiler that is not yet tested
+      ###         with Boost::Hana. Please tell us about whether it
+      ###         successfully compiles or if and how it failes.
+      ###
+      ###         Thanks !
+      ###
+      ")
+      SET(SHOW_PREFERRED_COMPILER ON)
+endif()
+
+if(SHOW_PREFERRED_COMPILER)
+    message("
+      ###
+      ###   ---   Consider switching to Clang >= 3.5.0   ---
+      ###
+      ###   If Clang is already installed on your system, then tell
+      ###   CMake about it by using
+      ###
+      ###     cmake -DCMAKE_CXX_COMPILER=clang++ ...
+      ###
+    ")
+endif()
+
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,8 @@
 # Distributed under the Boost Software License, Version 1.0.
 # (See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
 
+set(CMAKE_LEGACY_CYGWIN_WIN32 0)   # Prevent CMake showing a warning
+
 ##############################################################################
 # Setup project
 ##############################################################################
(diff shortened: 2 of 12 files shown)


cmake/CheckCxxCompilerCompatibility.cmake
```diff
@@ -0,0 +1,64 @@
+# Copyright Louis Dionne 2015
```
ldionne: You probably want your name there, not mine (x3).

ldionne:
Concerning the build failure: I broke the `develop` branch, and you (correctly) based your pull request on that broken commit. Can you rebase your branch on top of the latest `develop` so the tests pass? Sorry about that, it wasn't my best shot. It's late now; I'll have a look at the PR tomorrow and will leave some comments, but it looks good so far.

m-j-w:
I think I successfully rebased the pull request and added a contribution line. Sorry for the multiple commits, I'll squash it after your final code review. Good looking doesn't mean good behaving...

ldionne:
After looking at this PR quite extensively, here are my thoughts. There are actually three things we're trying to achieve here:

1. Add more user-friendly diagnostics for supported compilers during the CMake generation step. This is a great idea, and I like your take on it to have a separate file deal with it. One thing that worries me, however, is that some of these checks are already done in [boost/hana/config.hpp](https://github.com/ldionne/hana/blob/0ccb92776c8a488a8085c0a3f4ca4c8d23654ad8/include/boost/hana/config.hpp). I'm slightly worried about duplicating these checks such that they eventually get out of sync. Do you think it is not sufficient to have these checks in `hana/config.hpp`?
2. Add support for incomplete C++14 standard libraries. I'm not sure about this one. Hana has a very strict policy regarding compiler/stdlib requirements, as documented in the [README](https://github.com/ldionne/hana#hacking-on-hana). I am chilly at the idea of maintaining workarounds for non-C++14 stuff. It simply isn't worth it, since there are literally hundreds of workarounds we could do for specific platforms just to have to library's tests compile, but each of these workarounds makes our life more difficult in the long term (especially since they can't all be tested all of the time). Instead, my preferred approach would be to have Cygwin users install a C++14 library (e.g. update their libstdc++ or install libc++, as documented in the README). Is this too difficult/impossible on Cygwin, or is this a reasonable requirement?
3. Ensure the examples and unit tests build without support for exceptions, and at the same time test that the library itself does not use exceptions internally. This is a superb idea, and it is quite important because some potential clients of the library might not allow exceptions in their code. Also, as a metaprogramming library, it seems natural that exceptions should not be used in the implementation, and hence having a test configuration without exceptions seems desirable.

Before modifying this PR further, let's resolve this and decide what we want/do not want.

ldionne:
Actually, regarding support for building without exceptions, I even wonder whether that shouldn't be the default? Clearly, we don't want to rely on exceptions _anywhere_ in Hana, so it seems like it would be more stringent to put `-fno-exceptions` whenever the flag is supported.

m-j-w:
Checks should be performed depending on the systems default, whether with of without exceptions. In the end, it is a library to be embedded in another application. Though, limiting hana itself to not rely on exceptions appears plausible.

m-j-w:
@ldionne, although already closed, here my quick confirmation: Hana compiles well and succeeds with the tests as long as mentioned type traits and to_string are not required.

When including <boost/hana.hpp> and compiling some small dummy code, one receives the warning of C++14 not being provided. Luckily, this also is issued for gcc < 5.0 since then the __cplusplus equals 201300. Only clang pretends to fully comply with C++14 even when linking against libstdc++ < 5.0. 
So, looks as expected, looks good. Thanks !
___
Here the summary of the test build:

    $ cmake -DCMAKE_CXX_COMPILER=clang++ ..
    $ make -i check after cmake

resulted in

    99% tests passed, 13 tests failed out of 908
    Total Test time (real) =  19.36 sec
    The following tests FAILED:
              9 - example.any_of (Not Run)
             10 - example.ap (Not Run)
             57 - example.eval_if (Not Run)
            170 - example.overview (Not Run)
            215 - example.sequence.monad.types (Not Run)
            272 - example.tutorial.introspection.json (Not Run)
            274 - example.tutorial.quickstart (Not Run)
            277 - example.type.basic_type (Not Run)
            280 - example.type.integral (Not Run)
            295 - example.wandbox (Not Run)
            377 - test.header.boost.hana.ext.std.type_traits.hpp (Not Run)
            380 - test.header.boost.hana.ext.std.hpp (Not Run)
            776 - test.ext.std.type_traits (Not Run)
    Errors while running CTest
    [100%] Built target check

ldionne:
I checked all the failing tests/examples and I can understand why they would fail. However, I'm slightly surprised about `example.type.basic_type`. Is it because the `<cxxabi.h>` header is missing? I don't think this header is portable.

Also note that `example.any_of` and `example.type.integral` included `<boost/hana/ext/std/type_traits.hpp>`, without requiring it strictly speaking. I changed that and they should pass now.

m-j-w:
The current status with the reason of non-compilation / missing symbol is as follows, your conclusions are correct, the last one is new however:

    The following tests FAILED:
             10 - example.ap (Not Run)    =>    ap.cpp:20:9: error: static_assert expression is not an integral constant expression
             55 - example.eval_if (Not Run)    =>    std::is_trivially_copyable
            168 - example.overview (Not Run)    =>    std::is_trivially_copyable
            213 - example.sequence.monad.types (Not Run)    =>    std::is_trivially_copyable
            270 - example.tutorial.introspection.json (Not Run)    =>    std::to_string
            272 - example.tutorial.quickstart (Not Run)    =>    std::is_trivially_copyable
            275 - example.type.basic_type (Not Run)    =>     error: no member named 'free' in namespace 'std';  abi::__cxa_demangle(mangled, 0, 0, &status), std::free
            293 - example.wandbox (Not Run)    =>    std::to_string
            374 - test.header.boost.hana.ext.std.type_traits.hpp (Not Run)   =>   std::is_trivially_copyable
            377 - test.header.boost.hana.ext.std.hpp (Not Run)   =>   std::is_trivially_copyable
            750 - test.ext.std.type_traits (Not Run)   =>   std::is_trivially_copyable
            821 - test.pair.cnstr.copy (Not Run)    =>    std::is_trivially_copy_constructible
    Errors while running CTest
    [100%] Built target check

Anyway, I think it's fine, the important basic stuff is operational.

ldionne:
Ok, `example.type.basic_type` should be fixed for now, although it will have to be modified if we ever support MSVC (one can dream). Regarding `test.pair.cnstr.copy`, I just checked and it's also an expected failure (but the test was added just yesterday). Thanks a lot!

PR #170, opened 2015-09-01
operators::adl is already base class of basic_type< T >

bebuch:


ldionne:
This is going to fail on Clang. Agreed that I should try to understand what's going on instead of putting a band-aid by inheriting twice.

ldionne:
I admit being puzzled by the fact that this patch works. I could swear that I initially wrote it that way, saw it fail, and then cheaply worked around the failure by inheriting twice from `operators::adl`. Anyway, thanks; I will apply your commit locally to avoid a merge commit for such a small change.

ldionne:
Closing this, since 05a57e5 is already in develop.

Issue #201, opened 2015-11-09
[Build] Boost::Hana on Windows with Clang 3.7.0 using MinGW64 and Visual Studio 2015


m-j-w:
More a success story than issue: I've been building Hana with Clang 3.7.0 (libstdc++ of gcc 5.2.0), both provided by MinGW64, and it compiled and ran all tests and examples successfully (917 of 917), with and without exceptions enabled, respectively.

Secondly, I've tried LLVM/Clang on Windows with the prebuilt windows binaries from http://llvm.org/releases/download.html . After generating the solution with cmake, changing manually the platform toolset to `LLVM-vs2014`, I chose manually some of the tests and examples, and compiled and ran them successfully. This includes the known difficult tests with type traits and the 9 parts of tuple tests.
The build is a bit verbose due to some warnings about unrequired compiler flags, and features being GNU extensions, such like string literals. To my surprise, even the VS debugger works with the Clang-generated binaries... Unfortunately I'm not too familiar with the Visual Studio toolchain, so I can't provide a patch for the cmake scripts or to automate the testing framework. At least not in the near future.

Anyway, I'd say there are now at least two possible ways to use Boost::Hana on Windows. And Cygwin should also work fine once GCC is upgraded from 4.9.3 to 5.2.0 or higher.

badair:
This is great news. I'm going to try this tomorrow. Thanks!

ldionne:
@m-j-w  Thanks a bunch! I'll try to see if I can setup continuous integration on AppVeyor.

m-j-w:
Well, it appears to be rather trivial after all: in `cmd.exe`

    cd hana
    mkdir build
    cd build
    cmake .. -T LLVM-vs2014 -DCMAKE_CXX_COMPILER="C:\Program Files\LLVM\msbuild-bin\cl.exe" -G "Visual Studio 14 2015 Win64" -DCMAKE_CXX_FLAGS="-Wno-gnu-string-literal-operator-template"`

Then simply opening the created `Boost.Hana.sln` with Visual Studio 2015 and run from the menu `Tests > Run > All` and wait.. and wait... and wait... and __918 successful, 4 failed__.

The failing tests are:
- test.ext.std.array.part3 (hana\include\boost/hana/detail/algorithm.hpp(83,14): error : call to 'lexicographical_compare' is ambiguous)
- test.tuple.empty_member (hana\test\tuple\empty_member.cpp(17,9): error : static_assert failed)
[...]

PR #203, opened 2015-11-12
Compiler detection for LLVM/Clang with Visual Studio integration

m-j-w:
Provided are some changes to the compiler detection with cmake.
The invocation on the command line (cmd.exe) actually only requires for a 32bit build
```
cmake .. -TLLVM-vs2014
```
A 64bit build is generated with
```
cmake .. -TLLVM-vs2014 -G"Visual Studio 14 2015 Win64"
```
Other options should not be required on a clean cmake run. The additional comparisons are for the case when the initial invocation is missing the -T argument, and hence everything get's really messy. This happens in particular when an existing solution is changed afterwards, e.g. by invoking cmake again with other options, but might also happen by other manual changes in the solution and the automatically triggered cmake run.

The check for MSVC must be the first. Otherwise the MSC compiler will fail at the first `#warning`. MSC uses a `#pragma message` for the same purpose.
The standard library of MSVC can be checked by the compiler version `_MSC_VER` since its bundled. However, detecting MSVC is not sufficient; `_clang_` must also be present. Unfortunately, also this might not be enough since the compiler could be changed manually, hence the platform toolset must also start with `LLVM-vs...` to ensure the correct MSVC-internally provided configuration with the LLVM/Clang integration.


ldionne:
For reference, this pull-request is related to issue #201.

ldionne:
Once this passes, could you please squash the three first commits together (since they are just reworks one of another), and I'll merge this.

ldionne:
Oh yes, and if you could rebase on top of the current `develop` branch, it would be really nice.

m-j-w:
Hmm, that didn't work as expected...

ldionne:
If the shit hits the fan, just leave the commits as-is and I'll squash + rebase them as needed :-)

m-j-w:
Well, I should clean that up later. At some point I managed to push my squashed changes to my develop branch.
So just take that squashed commit db9daa69aff2c17cf37841f8d3b38644434cdfc0 on my develop branch. It should merge with your develop branch.

ldionne:
> Well, I should clean that up later. At some point I managed to push my squashed changes to my develop branch.

You want to push to your `windows_build` branch, not your `develop` branch. Pushing to your `windows_build` branch will update this pull request.

m-j-w:
Yeah, well, adding the + sign before the branch to enforce the push after squashing also helps a lot...

ldionne:
> Yeah, well, adding the + sign before the branch to enforce the push after squashing also helps a lot...

You can also use `git push -f` or `git push --force`

PR #205, opened 2015-11-12
[sort] Improve runtime and compile-time performance

ldionne:
This PR is an attempt to make `hana::sort` more efficient.


ldionne:
Closing because #246 is a better solution to this problem.

PR #218, opened 2015-12-07
[CMake] Add support for measuring code coverage.

ldionne:
This PR is just for exposition. After experimenting around, it turns out that code coverage results are not very useful because Hana is so full of templates. Basically, we trivially end up with nearly 100% coverage, because all the functions that are instantiated are indeed being run. Instead, something more useful would be a tool to check for uninstantiated templates.


ldionne:
The reason why I won't merge this is because it increases the maintenance burden without actually providing something useful. Instead, I'd rather just document the attempt and close this without merging.

PR #220, opened 2015-12-10
[experimental] printable optional

ricejasonf:
"Just" for kicks :stuck_out_tongue:

### DIFF
--- a/experimental/printable/optional.cpp
+++ b/experimental/printable/optional.cpp
@@ -1,0 +1,31 @@
+/*
+@copyright Jason Rice 2015
+Distributed under the Boost Software License, Version 1.0.
+(See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
+ */
+
+#include <boost/hana/assert.hpp>
+#include <boost/hana/experimental/printable.hpp>
+#include <boost/hana/integral_constant.hpp>
+#include <boost/hana/optional.hpp>
+
+#include <sstream>
+namespace hana = boost::hana;
+
+
+int main() {
+    {
+        std::ostringstream ss;
+        ss << hana::experimental::print(
+            hana::just(hana::int_c<5>)
+        );
+        BOOST_HANA_RUNTIME_CHECK(ss.str() == "just(5)");
+    }
+    {
+        std::ostringstream ss;
+        ss << hana::experimental::print(
+            hana::nothing
+        );
+        BOOST_HANA_RUNTIME_CHECK(ss.str() == "nothing");
+    }
+}
--- a/include/boost/hana/experimental/printable.hpp
+++ b/include/boost/hana/experimental/printable.hpp
@@ -25,6 +25,7 @@ Distributed under the Boost Software License, Version 1.0.
 
 // models for different containers
 #include <boost/hana/fwd/map.hpp>
+#include <boost/hana/fwd/optional.hpp>
 #include <boost/hana/fwd/set.hpp>
 #include <boost/hana/fwd/string.hpp>
 #include <boost/hana/fwd/type.hpp>
@@ -67,7 +68,7 @@ BOOST_HANA_NAMESPACE_BEGIN namespace experimental {
 #else
     struct print_t {
         template <typename T>
-        constexpr auto operator()(T const& t) const {
+        std::string operator()(T const& t) const {
             using Print = print_impl<typename hana::tag_of<T>::type>;
             return Print::apply(t);
         }
@@ -87,7 +88,7 @@ BOOST_HANA_NAMESPACE_BEGIN namespace experimental {
     template <typename S>
     struct print_impl<S, hana::when<hana::Sequence<S>::value>> {
         template <typename Xs>
-        static auto apply(Xs const& xs) {
+        static std::string apply(Xs const& xs) {
             std::string result = "(";
             auto comma_separated = hana::intersperse(xs, ", ");
             hana::for_each(comma_separated, [&result](auto const& x) {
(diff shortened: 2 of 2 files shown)


ricejasonf:
crap.. Sorry I didn't name the branch appropriately. I must also confess that it is only by the luck that the header files are alphabetized. :sweat_smile:

include/boost/hana/experimental/printable.hpp
```diff
@@ -111,6 +111,18 @@ BOOST_HANA_NAMESPACE_BEGIN namespace experimental {
         }
     };
 
+    // model for hana::optional
+    template <>
+    struct print_impl<hana::optional_tag> {
+        template <typename O>
+        static auto apply(O const& optional) {
+            return hana::maybe("nothing",
+                [](auto x) {
```
ldionne: This will make a copy of `x` because you take by value. Can you please use `auto const& x`?

include/boost/hana/experimental/printable.hpp
```diff
@@ -111,6 +111,18 @@ BOOST_HANA_NAMESPACE_BEGIN namespace experimental {
         }
     };
 
+    // model for hana::optional
+    template <>
+    struct print_impl<hana::optional_tag> {
+        template <typename O>
+        static auto apply(O const& optional) {
```
ldionne: When the optional is empty, you return `"nothing"`. Hence the return type will be deduced to be `char*` (or `char const*`, not sure), which is wrong. To make sure that we're returning a `std::string`, you could specify the return type as `std::string`. Or perhaps that was done on purpose?

include/boost/hana/experimental/printable.hpp
```diff
@@ -111,6 +111,18 @@ BOOST_HANA_NAMESPACE_BEGIN namespace experimental {
...
+            return hana::maybe("nothing",
+                [](auto x) {
+                    return "just " + hana::experimental::print(x);
```
ldionne: Instead of `"just x"`, I think it would be preferable to have `just(x)`. This way, if `x` is something complex, it might be easier to tell when the string representation of `x` finishes (because of the closing parenthesis). What do you think?

experimental/printable/optional.cpp
```diff
@@ -0,0 +1,32 @@
...
+(See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
+ */
+
+#include <boost/hana/assert.hpp>
+#include <boost/hana/experimental/printable.hpp>
+#include <boost/hana/integral_constant.hpp>
+#include <boost/hana/optional.hpp>
+
+#include <sstream>
+#include <string>
```
ldionne: Technically, this is not required because you don't use `std::string` directly.

experimental/printable/optional.cpp
```diff
@@ -0,0 +1,32 @@
+/*
+@copyright Louis Dionne 2015
```
ldionne: You probably want your name here instead of mine!

include/boost/hana/experimental/printable.hpp
ricejasonf: Perhaps they should all specify `std::string` as a return type, especially print_t::(). (I see it is specified in the doxygen version).

ricejasonf: Do you prefer I use `->` notation for the return type?

include/boost/hana/experimental/printable.hpp
ricejasonf: I was on the fence about it. I almost did `Just 5` to be like haskell, though I am not a haskell programmer.

ricejasonf:
I realized that I should probably also add an include to `fwd/optional.hpp` in `printable.hpp`.

I took the liberty of specifying the return type for `print_t::operator()` to `std::string` which requires the removal of `constexpr`, so I removed `constexpr` in the doxygen part as well. Let me know if that is a mistake.

I cleaned up the other stuff you mentioned and used `->` when specifying a return type (lmk if you don't like that). Should I create a new pull request with a single commit and a better branch name? 
`experimental/printable_optional` ??

include/boost/hana/experimental/printable.hpp
ldionne: Hmm, you're right; `hana::print` documents that it returns `std::string`, yet `print_impl::apply` does not enforce it. I guess the right thing to do is to specify the return type of all the `print_impl::apply`s to be `std::string`. Can you change that while you're at it? Also, I'd use normal return type syntax, since it's more consistent with the rest of the code. I rarely use trailing return type syntax elsewhere in the code (it just happened to be so).

ldionne:
> I realized that I should probably also add an include to fwd/optional.hpp in printable.hpp.

That's right, good catch.


> I took the liberty of specifying the return type for print_t::operator() to std::string which requires the removal of constexpr, so I removed constexpr in the doxygen part as well. Let me know if that is a mistake.

This is fine. I'm not sure about removing the `constexpr` keyword though, since it really says that the function object itself is constexpr, only its `operator()` isn't anymore. So technically, it's correct to leave `constexpr`.

Still, I think we should do as you suggest and remove it, since it will convey the proper intuition for most users. I think most users wouldn't see the situation the same way I do, and they would find it confusing.

> I cleaned up the other stuff you mentioned and used -> when specifying a return type (lmk if you don't like that). 

Good, but please use regular return type syntax. I'm not strongly attached to either (and in general I would probably pick the trailing syntax), but it's less consistent with the rest of the code.

> Should I create a new pull request with a single commit and a better branch name? experimental/printable_optional ??

No, just update your commit locally and `push --force` to the `printable_optional` branch of your own fork. That'll update this pull request.

ricejasonf:
Oh, now that you explain it... The function object itself IS constexpr. I was looking at the pseudo-lambda next the struct function object, which caused my confusion.

ldionne:
Looks good to me. When the CI is green, can you please squash these two commits together, write a simple but meaningful commit message for the added feature and then `push --force` to update this pull request with the single, squashed commit. I'll then merge.

ricejasonf:
I made a third commit changing the return type for all `print_impl`s. I only saw you mention that in a github email. I still don't see it in the converstation.

ldionne:
See [above](https://github.com/boostorg/hana/pull/220#discussion_r47262186). It's a comment on an _outdated diff_, as GitHub says. I can't say I'm a fan of GitHub's code review facilities, really.

ldionne:
Thank you!

PR #222, opened 2015-12-14
[bug] convert predicate results to bool

### LINKED ISSUE #221: [find_if] Predicates should be allowed to return an int_c.
I wanted to use hana::length as a predicate for `find_if` and `filter`, but it does not compile. Consider the following example.
```
#include<boost/hana.hpp>

namespace hana = boost::hana;

int main()
{
  static_assert(hana::find_if(hana::make_tuple(
      hana::make_tuple(),
      hana::make_tuple(),
      hana::make_tuple(hana::int_c<0>, hana::int_c<1>),
      hana::make_tuple()), hana::length)
        ==
      hana::just(hana::make_tuple(hana::int_c<0>, hana::int_c<1>)), "");
}
```

ricejasonf:
Fixes #221

### DIFF
--- a/include/boost/hana/count_if.hpp
+++ b/include/boost/hana/count_if.hpp
@@ -55,7 +55,7 @@ BOOST_HANA_NAMESPACE_BEGIN
             >::type>
             constexpr auto operator()(Xs&& ...xs) const {
                 constexpr bool results[] = {false, // <-- avoid empty array
-                    hana::value<decltype((*pred)(static_cast<Xs&&>(xs)))>()...
+                    static_cast<bool>(hana::value<decltype((*pred)(static_cast<Xs&&>(xs)))>())...
                 };
                 constexpr std::size_t total = detail::count(
                     results, results + sizeof(results), true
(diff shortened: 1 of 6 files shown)


ricejasonf:
I used `detail::decay` in partition.hpp because `Pred` was `const&`. I noticed in filter.hpp you use `Pred const& pred` with `pred` in the `decltype` which gets around that. Perhaps that is better idk

ricejasonf:
I would like to know why `partition` was working inside `sort`, but not in my test case. (before I used `detail::decay`)

ricejasonf:
because it was an lvalue?

ricejasonf:
Oh, I think it is because the return value is actually and rvalue because `id` forwards it's return value.

ldionne:
This looks good to me. Was there anything else you wanted to add to the PR?

ricejasonf:
No, thank you.

ricejasonf:
It says "boostorg:master". Can you change that?

ldionne:
Oops, seems like you created a PR against master. I can't change that on GitHub, but I'm going to merge the PR manually anyway.

Issue #223, opened 2015-12-17
[map] Optimize representation and operations


ldionne:
`hana::map` is currently fairly inefficient. It is implemented on top of `hana::tuple`, which is an index-based container. This causes `find_if`, `any_of`, `at_key`, `erase_key`, `insert` and friends to be implemented in terms of a linear search in the underlying tuple of pairs. This is unacceptably slow at compile-time.

By changing the representation of `hana::map`, we could most likely achieve interesting speedups. Consider for example the following representation (which is extremely to the underlying implementation of `hana::tuple`):

```c++
template <int index, typename Value>
struct holder { Value value; };

template <typename Indices, typename ...Pairs>
struct map_impl;

template <int ...index, typename ...Key, typename ...Value>
struct map_impl<std::integer_sequence<int, index...>, hana::pair<Key, Value>...>
    : holder<index, Value>...
{ };

template <typename ...Pairs>
struct map
    : map_impl<std::make_integer_sequence<int, sizeof...(Pairs)>, Pairs...>
{ };
```

Of course, we would need to generalize this to work for arbitrary `hana::Product`s, but it's a detail. Let's focus on `hana::pair`s for now. With this representation, we could implement `hana::find_if` as

```c++
template <typename Predicate, typename ...T>
constexpr int find_index_eager() {
    bool results[] = {
        decltype(std::declval<Predicate const&>()(std::declval<T const&>()))::value...
    };

    // Find first satisfied index
    for (int index = 0; index < static_cast<int>(sizeof...(T)); ++index) {
        if (results[index])
            return index;
    }

    return -1;
}

template <int index, typename Value>
constexpr auto get_helper(hana::int_<index>, holder<index, Value> const& element) {
    return hana::just(element.value);
}

constexpr auto get_helper(hana::int_<-1>, ...) {
    return hana::nothing;
}

template <typename ...Key, typename ...Value, typename Predicate>
constexpr auto find_if(map<hana::pair<Key, Value>...> const& m, Predicate const&) {
    constexpr int index = find_index_eager<Predicate, Key...>();
    return get_helper(hana::int_<index>{}, m);
}
```

This takes advantage of the fact that the order of elements in a `map` is not defined. Hence, we can eagerly compute the `Predicate` on all the keys of the map, which is most likely more efficient than doing a recursive linear scan (for small to average map sizes, which should be the most common). A similar technique could then be used to optimize various operations on a map.

[...]

PR #226, opened 2015-12-20
[map] Use an eager algorithm for map lookup

ricejasonf:
Related #223

### DIFF
--- a/include/boost/hana/map.hpp
+++ b/include/boost/hana/map.hpp
@@ -22,12 +22,14 @@ Distributed under the Boost Software License, Version 1.0.
 #include <boost/hana/concept/product.hpp>
 #include <boost/hana/config.hpp>
 #include <boost/hana/contains.hpp>
-#include <boost/hana/core/to.hpp>
+#include <boost/hana/core/is_a.hpp>
 #include <boost/hana/core/make.hpp>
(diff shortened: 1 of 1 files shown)


ricejasonf:
There may be more room for improvement. I'm not sure. After rebasing the benchmarks seemed to run .01-.02 seconds slower at n=400. 

Using the specialization for type_tag in `at_key` doesn't seem to make a difference so maybe I am doing something wrong there.

ricejasonf:
On my box, `benchmark.at_key.lookup.compile` shows map compiling in 1.38s for every test. It also reveals that `hana::set` scales linearly.

ricejasonf:
2911e959ffca7cda0dcac97e3b0f4823ce8308e2 removes some vestigal helpers I accidentally left in.

I tried having `make` create `map_impl` directly, but there was no noticeable performance difference.

ricejasonf:
For `benchmark.at_key.lookup.compile`, I substituted a lightweight Product for the map which puts it on par with tuple and set. (0.5s on my box)

ldionne:
OK, so I updated your branch with my modifications.

ldionne:
This is superseded by #247, which shows better compile-time performance.

PR #236, opened 2016-01-12
[experimental] printable hana::template_

ricejasonf:
If you would like, I can do this for the other `Metafunctions` for consistency, but first I wanted to gauge your reaction to giving them their own tags.

I think this would be useful for more than just `Printable`.

### DIFF
--- a/include/boost/hana/experimental/printable.hpp
+++ b/include/boost/hana/experimental/printable.hpp
@@ -33,6 +33,7 @@ Distributed under the Boost Software License, Version 1.0.
 #include <boost/core/demangle.hpp>
 
 #include <iostream>
+#include <regex>
 #include <sstream>
 #include <string>
 #include <typeinfo>
@@ -83,6 +84,12 @@ BOOST_HANA_NAMESPACE_BEGIN namespace experimental {
         using Tag = typename hana::tag_of<T>::type;
         static constexpr bool value = !hana::is_default<print_impl<Tag>>::value;
     };
+
+    namespace print_detail {
+        std::string strip_type_junk(std::string const& str) {
+            return std::regex_replace(str, std::regex("^([a-z_]+::)*([a-z_]*)_t<"), "$2<");
+        }
+    }
 
     // model for Sequences
     template <typename S>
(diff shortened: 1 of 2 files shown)


ricejasonf:
I made the suggested changes, but did not add any more implementaions. Perhaps the default should print the output from `boost::core::demangle`.

ldionne:
Random note: It's crazy and scary to think that any comment we made on the original diff disappears when we update the diff. Not-so-good job from GitHub on this.

ldionne:
Question: Why did you not implement it for other `Metafunction`s? Is it just a question of time, or is there any underlying issue that would prevent that?

ricejasonf:
>Random note: It's crazy and scary to think that any comment we made on the original diff disappears when we update the diff. Not-so-good job from GitHub on this.

I'd bet the information is still there, but the revision is no longer displayed in the ui.

>Question: Why did you not implement it for other Metafunctions? Is it just a question of time, or is there any underlying issue that would prevent that?

Actually I'm going to try doing a specialization for all `hana::Metafunction`s.

ricejasonf:
I tried implementing a `print_impl` with `when<hana::Metafunction<...>...`, but it was problematic getting it to match the variadic templates not to mention that `metafunction_class` has a different signature (ie it is not variadic).

I went ahead and added explicit specializations for each Metafunction and switched to using a **regex** to "prettyify" the output. If you don't want it pulling in `<regex>` I can change it back, but I think it is more robust this way.

ldionne:
That seems good. I don't mind the `<regex>` include. Can you please squash those two commits and I'll merge the PR?

ricejasonf:
I won't be getting to it immediately, but I would like to add Printable to the functions in Functional. Should merging be deferred and I add that here, or should I just do a new PR later?

ldionne:
Why do you need to print the functions in Functional? Printing a function properly is impossible in the general case.. In all cases, that should be attempted in a different PR.

ldionne:
Merged in 2b0ede886e922858e5c68b3732d52cbcad89a5c4.

ricejasonf:
>Why do you need to print the functions in Functional?

In my case I have trees with types and templates in them. It is useful to be able to use something like `hana::partial` on the result of `hana::template_` in some cases. So maybe it would look like:
```
partial(template<nbdl::StoreEmitter>, type<std::unordered_map>)
```
where `nbdl::StoreEmitter` is a template taking two types.

My current workaround has been to wrap the whole works in `hana::type_c<...>`, because otherwise `print` explodes.

ldionne:
Hmm. It would work fine when the function is a `template_`, but what about when the function is something arbitrary? For example, what would you expect `hana::experimental::print([]{})` to output?

ricejasonf:
lol idk. I should probably play around with it and see what is feasible.

ldionne:
Sure; don't hesitate to create another PR or an issue if you think this is a valuable addition, and we'll discuss. I'm just a bit skeptical, just as I would be if someone proposed to make functions `Comparable` (impossible in the general case). I'm not saying that printing functions is not feasible, but it certainly is not obvious how to do it sensibly.

Issue #118, opened 2015-06-14
[Associative Sequence] Consider using hash-based structures


ldionne:
See [this](http://article.gmane.org/gmane.comp.lib.boost.devel/261170) post on the mailing list.

ldionne:
Also have a look at [this](https://github.com/ldionne/hana/blob/2c1bb3a607dc03489d5f68df65f36cd65510d8dc/include/boost/hana/detail/sandbox/map.hpp) old implementation of a hash-based Map in the sandbox.

ricejasonf:
I was thinking that the point of **normalizing** the integral_constants was to make `m[size_c<5>] == m[int_c<5>]'. If the buckets contain a list of elements that are linearly searched using `Comparable` then it seems pointless to `int_c<5>` and `size_c<5>` in the same bucket. Is there ever a case where two `IntegralConstants` will be equal but have different types?

ricejasonf:
I would like to add some tests for the above when you get back to me on the desired behaviour. I would also add tests the check that `map` ultimately uses `Comparable`, but I argue that there is no realistic use case for that.

ldionne:
I will close this, as this is currently being considered in #242.

PR #241, opened 2016-01-20
[map] use type-level hash

ricejasonf:
- Hashable concept
    - hash function

### DIFF
--- a/include/boost/hana/concept/hashable.hpp
+++ b/include/boost/hana/concept/hashable.hpp
@@ -1,0 +1,25 @@
+/*!
+@file
+Defines `boost::hana::Hashable`.
+
+@copyright Jason Rice 2016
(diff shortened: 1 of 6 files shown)


ricejasonf:
As it relates to #118, I'm not sure if I hit the mark. Here, buckets hold exactly one element, `hash` only seems to accomplish converting integral constants to a common type, and `Comparable` is completely ignored.

Also, I may have gone overboard with the `Hashable` concept, but it seems that users should be able to specialize `hash_impl`.

There are some other rough edges that need rounding out, but I wanted to get some feedback on what I have so far.

ldionne:
Wtf? How did I close this?

ldionne:
Ahh! Please re-create this pull request, but targeting the `develop` branch instead!

PR #242, opened 2016-01-21
[map] Use a hash-based approach for lookup

ricejasonf:
Note that this is branched off of my original PR #226.

### DIFF
--- a/include/boost/hana/concept/hashable.hpp
+++ b/include/boost/hana/concept/hashable.hpp
@@ -1,0 +1,25 @@
+/*!
+@file
+Defines `boost::hana::Hashable`.
+
+@copyright Jason Rice 2016
+Distributed under the Boost Software License, Version 1.0.
+(See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
+ */
+
+#ifndef BOOST_HANA_CONCEPT_HASHABLE_HPP
+#define BOOST_HANA_CONCEPT_HASHABLE_HPP
+
+#include <boost/hana/fwd/concept/hashable.hpp>
+
+#include <boost/hana/config.hpp>
+
+
+BOOST_HANA_NAMESPACE_BEGIN
+    template <typename X>
+    struct Hashable {
+        static constexpr bool value = true;
+    };
+BOOST_HANA_NAMESPACE_END
+
+#endif // !BOOST_HANA_CONCEPT_HASHABLE_HPP
--- a/include/boost/hana/detail/hash_table.hpp
+++ b/include/boost/hana/detail/hash_table.hpp
@@ -1,0 +1,215 @@
+/*!
+@file
+Defines `boost::hana::detail::hash_table`.
+
+@copyright Jason Rice 2016
+Distributed under the Boost Software License, Version 1.0.
+(See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
+ */
+
+#ifndef BOOST_HANA_DETAIL_HASH_TABLE_HPP
+#define BOOST_HANA_DETAIL_HASH_TABLE_HPP
+
+#include <boost/hana/detail/index_if.hpp>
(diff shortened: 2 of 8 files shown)


ricejasonf:
Sorry I'm going to close this. It was a lot simpler to make new, appropriately names branch off of `benchmarks/map` and cherry-pick the commit from this. I can do a new PR when `make check` is complete.

I would like to know if I even came close to what you guys were talking about regarding the type-level hash, which was the intent of this PR.

ldionne:
> Sorry I'm going to close this
Why close this?



> I would like to know if I even came close to what you guys were talking about regarding the type-level hash, which was the intent of this PR.

Yes, this is very close, except that if two hashes collide then we'll get the wrong results. Each bucket has to contain a list of elements whose hash is that of the bucket, and then we'll do a linear search inside the bucket. Or you could use basically any hashing scheme (linear probing or something more exotic), if you wish.

ricejasonf:
I was thinking that the point of normalizing the integral constants was to make

m[size_c<5>] == m[int_c<5>]

If the buckets contain a list of elements that are linearly searched using Comparable then it seems pointless to put int_c<5> and size_c<5> in the same bucket. Is there ever a case where two IntegralConstants will be equal but have different types?

I would like to add some tests for the above when you get back to me on the desired behaviour. I would also add tests to check that map ultimately uses Comparable, but I argue that there is no realistic use that justifies it.

ldionne:
> I was thinking that the point of normalizing the integral constants was to make
> 
> ```c++
> m[size_c<5>] == m[int_c<5>]
> ```

Correct. More specifically, the requirement is that if `hana::equal(x, y)`, then `hana::equal(hash(x), hash(y))`. This ensures that we fall into the right bucket all the time, since the hash of a value determines the bucket in which it will end up in the map.


> If the buckets contain a list of elements that are linearly searched using Comparable then it seems pointless to put int_c<5> and size_c<5> in the same bucket. 

Looking up in such a map is a two-step process. First, we compute the hash of the key, which gives us the bucket. Now, many different keys can have the same hash, and hence fall into the same bucket, since the hash is not expected to be perfect (collision free). Once we've found the bucket, we then do a linear search __within__ this bucket to find the key that we're interested in. If `hash(int_c<5>) != hash(long_c<5>)`, then they will fall into different buckets, and we'll never find `int_c<5>` if we search for `long_c<5>` (and vice versa).


> Is there ever a case where two IntegralConstants will be equal but have different types?

Yes, for example `int_c<5> == long_c<5>`, but `decltype(int_c<5>) != decltype(long_c<5>)`.

ricejasonf:
`ef2f06f` I made the buckets contain indices of map's storage a tuple.  With regard to `benchmark.at_key.lookup`, there is a greater upfront cost, but lookups appear to scale better (4s vs 21s @n=200). `find` and `erase_key` use the key based lookup so I expect they will perform better as well. The other functions should be unchanged from master as the map's storage is once again a `tuple`.

ricejasonf:
BTW, `detail::hash_table` could easily be made the back-end for `hana::set` as well.

ricejasonf:
For my particular use case I have a better way than trying to use sets as keys. Since sets or other containers as keys would probably always result in a linear search, then perhaps it is better to delete the default implementation for `hash` so if people try to use something as a key they know what they are getting into.

ricejasonf:
I tried using `all_of` to check if all of the keys were "fast", meaning they could never have a hash collision. It shaves about 10% off of the `at_key.index_of_lookup` benchmark, but loses the ability to check for duplicates at creation time. I won't bother pushing, but I thought it was worth mentioning.

ricejasonf:
I just pushed `295f5fc` to fix `hash` for `type_tag`.

ldionne:
Superseded by #247, where latest changes on the Hashable have been made. #247 also shown a small compile-time speedup over this PR, due to different techniques being used.

PR #246, opened 2016-01-28
[sort] Use a stable insertion sort

ldionne:
This is related to #205. Preliminary benchmarking seems to indicate that this insertion sort technique is __way__ better than the eager sorting done in #205.


ldionne:
Here is a benchmark of the different sorting techniques. The _original quicksort_ represents the current sort, the _eager sort_ represents #205, and the two _insertion sorts_ are this PR. 

screen shot 2016-01-28 at 11 30 48

Issue #213, opened 2015-11-21
Document operators defined for concepts.


ricejasonf:
In `include/boost/hana/detail/operators/*.hpp` there are operators defined for concepts that aren't making it into the documentation. They don't show up in the search bar, and I was unable to find them in side-menu including `Details -> adl`.

One example is the `operator|` for the Monad `chain` function.

ldionne:
Actually, these operators are defined per-datatype. Indeed, they are not provided by the concept, but by the container. For example, see [hana::tuple's documentation](http://boostorg.github.io/hana/structboost_1_1hana_1_1tuple.html). There's a section called _provided operators_.

ricejasonf:
Aha. I noticed some operators show up in the side menu and the search bar but not others.

ldionne:
Which ones?

You know what; I think we can do better by actually documenting `operator|` & friends with Doxygen for each container. Something like

```c++
#ifdef BOOST_HANA_DOXYGEN_INVOKED
//! Equivalent to `hana::chain`.
template <typename ...T, typename F>
constexpr auto operator|(tuple<T...>, F);
#endif
```

ldionne:
This is fixed by PR #250.

PR #250, opened 2016-02-03
Documentation/operators

ricejasonf:
There are two things to note here:

1. I took the liberty of removing documentation of a pointer operand for `integral_constant` because I think it no longer exists. It is in a separate commit just in case.

2. For the data-types that I added a documentation definition to, for some reason the "Public Members" documentation is not full width, and I could not find any reason just by looking at the style rules. I could possibly add a style rule unless I am missing something obvious with this.

### DIFF
--- a/include/boost/hana/fwd/lazy.hpp
+++ b/include/boost/hana/fwd/lazy.hpp
@@ -68,7 +68,11 @@ BOOST_HANA_NAMESPACE_BEGIN
     //! [1]: http://ldionne.com/2015/03/16/laziness-as-a-comonad
 #ifdef BOOST_HANA_DOXYGEN_INVOKED
     template <typename ...>
-    struct lazy;
+    struct lazy {
+        //! Equivalent to `hana::chain`.
+        template <typename ...T, typename F>
+        friend constexpr auto operator|(lazy<T...>, F);
+    };
 #else
     // We do not _actually_ define the lazy<...> type. Per the documentation,
     // users can't rely on it being anything, and so they should never use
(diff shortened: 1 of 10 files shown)


include/boost/hana/fwd/lazy.hpp
```diff
@@ -68,7 +68,11 @@ BOOST_HANA_NAMESPACE_BEGIN
     //! [1]: http://ldionne.com/2015/03/16/laziness-as-a-comonad
 #ifdef BOOST_HANA_DOXYGEN_INVOKED
     template <typename ...>
-    struct lazy;
+    struct lazy {
+        //! Equivalent to `hana::chain`.
+        template <typename ...T, typename F>
+        constexpr auto operator|(lazy<T...>, F);
```
ldionne: This should probably be made a `friend`, otherwise it does not make sense to receive two arguments. The same applies for some operators in the other files you modified.

ricejasonf: I just broke all of these free functions out into a separate file and then realized that maybe you meant that I simply add the `friend` keyword there. Which is preferred?

ricejasonf:
I just broke all of these free functions out into a separate file and then realized that maybe you meant that I simply add the friend keyword there. Which is preferred?

ricejasonf:
I perhaps went too far with breaking out the free operators, but I pushed it for review to see if this is a better solution. If not, I can reset back to what I had before and work from there. lmk

ldionne:
I really just meant that you should stick `friend` in front of the functions :p.

ricejasonf:
I reset back to earlier and simply added `friend` to all free operators. I also added `type`'s unary+ and everything for `integral_constant`.

ldionne:
That looks perfect. I will merge as soon as the CI is finished. Thanks a lot!

ricejasonf:
btw #213

ricejasonf:
No sweat! :sweat_smile:

ldionne:
Closed by bae4dc1d8dec14a17e6c8c82c436e723dc7a4514.

PR #253, opened 2016-02-19
Correctly forward std::extent value_type to hana::traits::extent

### LINKED ISSUE #252: Wrong return type of hana::traits::extent
Hana's extent wrapper returns integral_constant of "const size_t" type instead of "size_t" as other traits do.

So, this code
```cpp
std::cout << boost::core::demangle(typeid(
    hana::traits::extent(hana::type_c<int[2]>)
).name()) << std::endl;
```
outputs
```
boost::hana::integral_constant<unsigned long const, 2ul>
```

This results to unexpected error when doing something like
```cpp
hana::product<hana::integral_constant_tag<size_t>>(extents)
```
We need to specify Ring type here to handle empty extents but it doesn't compiles because "size_t" and "const size_t" treated as different Rings.

snizovtsev:
Fixes #252

### DIFF
--- a/include/boost/hana/traits.hpp
+++ b/include/boost/hana/traits.hpp
@@ -117,10 +117,10 @@ BOOST_HANA_NAMESPACE_BEGIN namespace traits {
         template <typename T, typename N>
         constexpr auto operator()(T&&, N const&) const {
             constexpr unsigned n = N::value;
-            constexpr auto result = std::extent<
+            using Result = std::extent<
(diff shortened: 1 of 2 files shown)


ldionne:
Thanks for the bug report and the PR! Everything looks fine apart from the nitpicks I commented in ecacde6. Once these are addressed, I'll merge your PR as soon as the CI is finished.

ldionne:
Thank you!

Issue #260, opened 2016-03-08
`hana::reverse_partial` has no default constructor.


ricejasonf:
Since `hana::partial` has a default constructor, perhaps it would be consistent to provide one for `hana::reverse_partial` as well.

My use case is partially applied `hana::Metafunction`s. It would also be cool to have partially applied Metafunction's that are themselves `hana::Metafunction`s. It could be implemented with `hana::experimental::types`. Maybe it could be called `hana::partial_mf` or something. I am not sure about the naming.

I can do a PR for one or both, if you are interested.

ldionne:
I didn't even know that `hana::partial` was default constructible. IOW, it's probably a product of the implementation but something officially provided. But I can sort of see why this might be useful. If we decide that `hana::partial` should be default-constructible, then of course `hana::reverse_partial` should be too. Can you open a separate issue for `hana::partial` metafunctions?

PR #261, opened 2016-03-08
[issue] add default cnstr to reverse_partial_t

ricejasonf:
This PR is to help with #260.

Note that I copied the comment from the implementation of `partial_t`, though I wouldn't expect a default constructor to be implicitly provided in the presence of a user-declared constructor. I'm not really sure what it is saying.


ldionne:
@ricejasonf I added my own test case instead of yours, which was more focused on metafunctions than default-constructibility.

PR #282, opened 2016-05-27
[overload_linearly] Allow `overload_linearly` to be called with a single argument

### LINKED ISSUE #280: Allow `overload_linearly` to be called with a single argument
Currently, [`overload_linearly`](http://www.boost.org/doc/libs/1_61_0/libs/hana/doc/html/group__group-functional.html#gaa46de6f618d9f14edb1589b36b6e75ec) requires at least two arguments. [`overload`](http://www.boost.org/doc/libs/1_61_0/libs/hana/doc/html/group__group-functional.html#ga83e71bae315e299f9f5f9de77b012139) can be successfully called with a single argument.

Apart from the inconsistency between the two functions, it's also very useful to allow `overload_linear` to be called with a single argument in generic contexts where the user is passing a variadic number of functions:

```cpp
template <typename... TFs>
void called_by_user(TFs&&... fs)
{
    return hana::overload_linearly(some_wrapper(FWD(fs))...);
}

called_by_user([](auto){ }); // error!
```

SuperV1234:
Fixes #280.

> [overload_linearly] Allow `overload_linearly` to be called with a single argument
>
> - Added an `make_overload_linearly_t::operator()` overload that takes a single argument, and simply > returns the argument

*(I closed the previous pull request because I completely forgot to create a new branch.)*

### DIFF
--- a/include/boost/hana/functional/overload_linearly.hpp
+++ b/include/boost/hana/functional/overload_linearly.hpp
@@ -11,6 +11,7 @@ Distributed under the Boost Software License, Version 1.0.
 #define BOOST_HANA_FUNCTIONAL_OVERLOAD_LINEARLY_HPP
 
 #include <boost/hana/config.hpp>
+#include <boost/hana/detail/decay.hpp>
 
 #include <type_traits>
 #include <utility>
@@ -98,7 +99,7 @@ BOOST_HANA_NAMESPACE_BEGIN
         }
 
         template <typename F>
-        constexpr decltype(auto) operator()(F&& f) const {
+        constexpr typename detail::decay<F>::type operator()(F&& f) const {
             return static_cast<F&&>(f);
         }
     };
--- a/include/boost/hana/functional/overload_linearly.hpp
+++ b/include/boost/hana/functional/overload_linearly.hpp
@@ -96,6 +96,11 @@ BOOST_HANA_NAMESPACE_BEGIN
             return (*this)(static_cast<F&&>(f),
                     (*this)(static_cast<G&&>(g), static_cast<H&&>(h)...));
         }
+
+        template <typename F>
+        constexpr decltype(auto) operator()(F&& f) const {
+            return static_cast<F&&>(f);
+        }
     };
 
     constexpr make_overload_linearly_t overload_linearly{};
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -54,12 +54,9 @@ install:
   - cmake --version
 
   ############################################################################
-  # Install a pre-built Clang-cl
+  # Setup the path to Clang-cl
   ############################################################################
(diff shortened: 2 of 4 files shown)


include/boost/hana/functional/overload_linearly.hpp
```diff
@@ -96,6 +96,11 @@ BOOST_HANA_NAMESPACE_BEGIN
             return (*this)(static_cast<F&&>(f),
                     (*this)(static_cast<G&&>(g), static_cast<H&&>(h)...));
         }
+
+        template <typename F>
+        constexpr decltype(auto) operator()(F&& f) const {
```
ldionne: If you call `overload_linearly(some-expr)` where `some-expr` is a temporary, which should often be the case because we'll use it with a lambda like `overload_linearly([](...) {...})`, you'll end up returning an rvalue reference to the temporary. Instead, and to mimic the behaviour of `overload`, you should use `hana::detail::decay` (equivalent to `std::decay`, but faster) to make sure a copy of the function is made.

ldionne: What I mean is that instead of `decltype(auto)`, it should be `typename detail::decay<F>::type` according to my argument above (which you could argue against though).

SuperV1234:
I agree with your argument, @ldionne: the updated commit (8e0c397ea9988d834a0f85f1da732bdc7124668d) now returns `typename detail::decay<F>::type`.

Two unrelated nitpicks:

* There are many instances in the codebase where `std::decay` is used instead of `detail::decay`. Is there any reason for that? 

* There also are many instances where `typename detail::decay<...>::type` is used - is there any reason for not defining a `decay_t` alias to avoid the `typename [...] ::type` boilerplate?

I could work on fixing both points if they're oversights.

ldionne:
Thanks, that looks perfect. Regarding the nitpicks:

> There are many instances in the codebase where std::decay is used instead of detail::decay. Is there any reason for that?

Oversight and laziness. But `detail::decay` should already be used in most places that are "performance critical". I would welcome a PR using `detail::decay` instead of `std::decay` in implementation details, but don't use `detail::decay` in examples and unit tests; it's just an implementation detail.


> There also are many instances where `typename detail::decay<...>::type` is used - is there any reason for not defining a `decay_t` alias to avoid the `typename [...] ::type` boilerplate?

As an implementation detail for optimization purposes, I'd rather not add an additional indirection through a template alias. It's probably negligible, but the added readability is also negligible IMO (since `detail::decay` is already being used only in a few, narrow places.

ldionne:
Can you please rebase on top of the current develop? That'll allow the Travis build to pass and then I'll merge.

SuperV1234:
Done.

ldionne:
Closed by 03b948918e0935ee0255ab8c442141b2ec175467...a5989ca1901f79b6a0d315ad764f28b81ec0a4cb

PR #284, opened 2016-05-30
Use detail::decay instead of std::decay in implementation

SuperV1234:
* In implementation code, occurrences of `typename std::decay<...>::type` have been replaced with `typename hana::detail::decay<...>::type`.

* The rationale for this change, as mentioned in #282, is that `hana::detail::decay` performs better than `std::decay` in "libc++". I also think that using the same `decay` implementation throughout the library is more consistent.

### DIFF
--- a/include/boost/hana/comparing.hpp
+++ b/include/boost/hana/comparing.hpp
@@ -12,6 +12,7 @@ Distributed under the Boost Software License, Version 1.0.
 
 #include <boost/hana/fwd/comparing.hpp>
 
+#include <boost/hana/detail/decay.hpp>
 #include <boost/hana/config.hpp>
 #include <boost/hana/equal.hpp>
 
@@ -37,7 +38,7 @@ BOOST_HANA_NAMESPACE_BEGIN
     //! @cond
     template <typename F>
     constexpr auto comparing_t::operator()(F&& f) const {
-        return detail::equal_by<typename std::decay<F>::type>{static_cast<F&&>(f)};
+        return detail::equal_by<typename detail::decay<F>::type>{static_cast<F&&>(f)};
     }
     //! @endcond
 BOOST_HANA_NAMESPACE_END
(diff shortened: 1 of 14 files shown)


include/boost/hana/comparing.hpp
```diff
@@ -12,6 +12,7 @@ Distributed under the Boost Software License, Version 1.0.
 
 #include <boost/hana/fwd/comparing.hpp>
 
+#include <boost/hana/detail/decay.hpp>
```
ldionne: Can you please sort the `#includes` alphabetically? This is really a small detail, but I try to keep them sorted as it makes it easier to detect duplicates and similar errors.

include/boost/hana/fuse.hpp
```diff
@@ -12,6 +12,7 @@ Distributed under the Boost Software License, Version 1.0.
 
 #include <boost/hana/fwd/fuse.hpp>
 
+#include <boost/hana/detail/decay.hpp>
```
ldionne: Same here, please sort.

SuperV1234:
Done (assuming you still want `/fwd/...` includes to appear first.)

ldionne:
Yeah, that's perfect. Thanks a lot!

ldionne:
Ah, I stumbled on another thing. It sometimes is possible to remove `<type_traits>` includes since nothing in it is used anymore. That's the case at least for `comparing.hpp`. Can you have a look? I know these are all details, but I think that keeping a code base clean is a matter of these.

SuperV1234:
I've removed unnecessary `#include <type_traits>` directives in the files where I replaced `std::decay` with `detail::decay`.

ldionne:
Please rebase on top of `develop` and this should pass the CI (I fixed the Travis build). Sorry for the annoyance.

ldionne:
Oh, and while you're at it, it would be great if you could squash these three commits into one, since that's really just one logical change set.

ldionne:
Nevermind, I merged manually. Thanks a lot for the patch!

PR #291, opened 2016-06-21
[Travis] Cache various downloads

ldionne:


ldionne:
Will close this PR because it's super long to test, and it's kinda stupid to cache things that we don't have to build anyway. I'll still merge the LLVM caching part, though.

Issue #290, opened 2016-06-19
Accessing map with result of `decltype_` does not compile


jacquelinekay:
I am trying to index into a `hana::map` using the result of `hana::decltype_` and iterating over the contents of a tuple.

My understanding was that `hana::decltype_` produces a `hana::type_c`, which seemed to be suitable for indexing into my map. I have also tried constructing a key using language decltype and wrapping that in `type_c`, but I got the same error.

Is this incorrect usage?

```
#include <iostream>
#include <boost/hana.hpp>

namespace hana = boost::hana;

enum Id : uint8_t {
  a, b, c
};

constexpr auto id_map = hana::make_map(
    hana::make_pair(hana::type_c<std::string>, Id::a),
    hana::make_pair(hana::type_c<int>, Id::b),
    hana::make_pair(hana::type_c<bool>, Id::c)
  );

int main() {
  hana::tuple<std::string, int, bool> types_tuple;

  hana::for_each(types_tuple, [](const auto & x) {
    std::cout << "ID is: " << id_map[hana::decltype_(x)] << std::endl;
  });
}
```

Gives this error in clang 3.8 (truncated significantly), on current master:

```
/home/jackie/code/cmbml/external/hana/include/boost/hana/optional.hpp:108:9: error: static_assert
      failed "hana::optional::operator* requires a non-empty optional"
        static_assert(detail::wrong<dummy...>{},
        ^             ~~~~~~~~~~~~~~~~~~~~~~~~~
/home/jackie/code/cmbml/external/hana/include/boost/hana/map.hpp:358:52: note: in instantiation of
      function template specialization 'boost::hana::optional<>::operator*<>' requested here
            constexpr std::size_t index = decltype(*MaybeIndex{}){}();
                                                   ^
/home/jackie/code/cmbml/external/hana/include/boost/hana/at_key.hpp:49:23: note: in instantiation
      of function template specialization 'boost::hana::at_key_impl<boost::hana::map_tag,
      void>::apply<const
      boost::hana::map<boost::hana::detail::hash_table<boost::hana::detail::bucket<std::__cxx11::basic_string<char>,
      0>, boost::hana::detail::bucket<int, 1>, boost::hana::detail::bucket<bool, 2> >,
      boost::hana::basic_tuple<boost::hana::pair<boost::hana::type_impl<std::__cxx11::basic_string<char>
      >::_, Id>, boost::hana::pair<boost::hana::type_impl<int>::_, Id>,
[...]

PR #293, opened 2016-07-08
[type] Add hana::typeid_ and deprecate hana::decltype_

ldionne:
Candidate for satisfying #290. I'm still unsure whether using `typeof` would be better, though.

### DIFF
--- a/include/boost/hana/fwd/type.hpp
+++ b/include/boost/hana/fwd/type.hpp
@@ -121,6 +121,11 @@ BOOST_HANA_NAMESPACE_BEGIN
     //! `decltype` keyword, lifted to Hana.
     //! @relates hana::type
     //!
+    //! @deprecated
+    //! The semantics of `decltype_` are can be confusing, and `hana::typeid_`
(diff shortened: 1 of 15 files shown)


include/boost/hana/fwd/type.hpp
```diff
@@ -121,6 +121,11 @@ BOOST_HANA_NAMESPACE_BEGIN
     //! `decltype` keyword, lifted to Hana.
     //! @relates hana::type
     //!
+    //! @deprecated
+    //! The semantics of `decltype_` are can be confusing, and `hana::typeid_`
```
viboes: s/are can/can/

ldionne: Thanks!

ricejasonf:
`typeof` :+1:

ldionne:
I decided against `typeof`, even though I prefer it, because it clashes with a GCC extension of the same name. When compiling with `std=gnu++14`, using `typeof` anywhere in your program causes a compilation error.

PR #294, opened 2016-07-08
Unblock testing with MSVC

MarcelRaad:
MSVC doesn't like the #warning directive and issues a fatal error.
It also sets _MSVC_LANG to 201402 for C++14 instead of __cplusplus, but that's the minimum value anyway.
It also doesn't like using {} instead of ::value for some type traits sometimes and issues an error about a missing default constructor. ::value is used most of the time anyway.

### DIFF
--- a/include/boost/hana/config.hpp
+++ b/include/boost/hana/config.hpp
@@ -71,7 +71,11 @@ Distributed under the Boost Software License, Version 1.0.
 // Check the compiler for general C++14 capabilities
 //////////////////////////////////////////////////////////////////////////////
 #if (__cplusplus < 201400)
-#   warning "Your compiler doesn't provide C++14 or higher capabilities. Try adding the compiler flag '-std=c++14' or '-std=c++1y'."
+#   if defined(_MSC_VER)
+#       pragma message("Warning: Your compiler doesn't provide C++14 or higher capabilities. Try adding the compiler flag '-std=c++14' or '-std=c++1y'.")
+#   else
+#       warning "Your compiler doesn't provide C++14 or higher capabilities. Try adding the compiler flag '-std=c++14' or '-std=c++1y'."
+#   endif
 #endif
 
 //////////////////////////////////////////////////////////////////////////////
--- a/example/detail/wrong.cpp
+++ b/example/detail/wrong.cpp
@@ -12,7 +12,7 @@ struct base_template {
     // static_assert(false, "...");
 
     // So instead we write this:
-    static_assert(hana::detail::wrong<base_template<T, U>>{},
+    static_assert(hana::detail::wrong<base_template<T, U>>::value,
     "base_template does not have a valid default definition");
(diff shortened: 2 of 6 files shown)


ldionne:
Do you mean that Hana's tests actually work on MSVC with this pull request? Which version of MSVC are you using?

MarcelRaad:
A few work, but unfortunately most of them don't. That's at least blocked by C++14 constexpr and a few bugs related to decltype in MSVC 14.3.

But it's a bit unfortunate that all of them currently fail because of the #warning directive rather than compiler bugs as MSVC could be fully standards-conformant in a few months (see for example Andrew Pardoe's comments in https://www.reddit.com/r/cpp/comments/4r8n41/what_are_some_reasons_why_visual_studio_seems_so/).

include/boost/hana/config.hpp
```diff
@@ -70,7 +70,7 @@ Distributed under the Boost Software License, Version 1.0.
 //////////////////////////////////////////////////////////////////////////////
 // Check the compiler for general C++14 capabilities
 //////////////////////////////////////////////////////////////////////////////
-#if (__cplusplus < 201400)
+#if (__cplusplus < 201400) && (defined(__clang__) || !defined(_MSC_VER))
```
ldionne: Why not simply `#if (__cplusplus < 201400) && !defined(_MSC_VER)`? And when `#warning` is not supported, we should fallback to `#pragma message` instead. Giving this message is important to avoid users being totally confused by meaningless error messages when trying to use Hana with a non-compliant compiler.

MarcelRaad: > Why not simply #if (__cplusplus < 201400) && !defined(_MSC_VER) ?

Because I didn't know that clang in MSVC mode also supports `#pragma message`. But it does, just tested with clang/c2, so I've updated to use `#pragma message` in MSVC mode.

ldionne:
By any chance, do you think it would be possible to setup Appveyor builds for MSVC (while allowing failures) so that we could track down its progression?

ldionne:
It would also make this PR more concretely useful, because right now I personally have no way to check the correctness/usefulness of the PR; I can only blindly trust that it makes things better on MSVC.

MarcelRaad:
> By any chance, do you think it would be possible to setup Appveyor builds for MSVC (while allowing failures) so that we could track down its progression?

> It would also make this PR more concretely useful, because right now I personally have no way to check the correctness/usefulness of the PR; I can only blindly trust that it makes things better on MSVC.

I fully agree. Unfortunately I have no idea about configuring AppVeyor builds using build scripts instead of Visual Studio project files, sorry :-(

ldionne:
Merged in 1b005c5679f5547dcdea3342dcb47ea5c54b9436, sorry for the long wait.

MarcelRaad:
Thanks!

PR #301, opened 2016-09-28
Official support for views

ldionne:
The goal of this PR is to officially support views, which are only experimental right now.

Still left to do:
- [ ] Move view types out of the `experimental` namespace, perhaps into `detail`
- [ ] Consider providing access to the `flattened` methods & al.
- [ ] Test nested views
- [ ] Add examples
- [ ] Document views in the tutorial (requires some rewording of the sections on containers and on algorithms)
- [ ] Provide operators
- [ ] Provide a non-dummy implementation of `hana::equal` and `hana::ap`
- [ ] Consider providing a `filtered_view`


ricejasonf:
I have a PR for `equal`, and I'd like to do `ap` as well. It appears that `detail::cartesian_product_indices` provides and interface that should make that easy to implement. A `cartesian_product` view looks pretty easy with that as well.

ricejasonf:
For `cartesian_product`, I think all I need now are tests. Check it out:

https://github.com/boostorg/hana/compare/pr/views...ricejasonf:view/cartesian_product?expand=1

Let me know what you think, and I will move forward with the tests sometime after the weekend.

ldionne:
I'm striking a few things that I marked as being required to ship this PR. Given my lack of time to work on Hana (very sadly), I'm going to go for a minimal implementation for the sake of shipping this. This PR is important and it has been sitting around for way too long.

ricejasonf:
I didn't actually do `ap` yet, but it should be trivial now.

I could help with tests too if that is welcome.

ldionne:
I think the tests for `ap` are okay, what I'm mostly concerned about is tests for the rest of the functionality. For example, `filter` _should_ work currently because `view_tag` implements `MonadPlus`, but I'm fairly confident it won't work properly (either compilation error or dangling reference) based on looking at how `single_view` is implemented.

So yeah, I'd welcome help on making the tests more robust. I'm happy to tackle the documentation side.

ricejasonf:
Are we going to make the views `Sequence` or is that some kind of violation?

ldionne:
They can't be `Sequence`s because there's no way to `make` them.

Issue #304, opened 2016-10-19
std::array adaptor fails for movable but non-copyable types


iolojz:
Attempting to turn a rvalue of type `std::array<Foo, N>` into a `boost::hana::tuple<...>` when `Foo` is non-copyable but movable does not work.  e.g.

    #include <boost/hana.hpp>
    #include <boost/hana/ext/std/array.hpp>
    
    struct Foo
    {
        Foo( void ) = default;
        Foo( const Foo & ) = delete;
        Foo( Foo && ) = default;
    };
    
    using bar = decltype(boost::hana::to_tuple( std::array<Foo, 42>() ));
    
    int main( void )
    {
        return 0;
    }

does not compile.

ricejasonf:
Hana is using `operator[]` to get the elements from the array which apparently doesn't not support rvalues. Using `std::get<n>` fixes it. I'll submit a PR momentarily.

ldionne:
Fixed by #305.

PR #305, opened 2016-10-22
Fix Github Issue #304

ricejasonf:
- Apparently std::array operator[] returns an lvalue reference when unpacking
    a temporary std::array. Using std::get<n> returns the
    proper reference type in the case of rvalues.

### DIFF
--- a/include/boost/hana/ext/std/array.hpp
+++ b/include/boost/hana/ext/std/array.hpp
@@ -93,7 +93,7 @@ BOOST_HANA_NAMESPACE_BEGIN
         template <typename Xs, typename N>
         static constexpr decltype(auto) apply(Xs&& xs, N const&) {
             constexpr std::size_t n = N::value;
-            return static_cast<Xs&&>(xs)[n];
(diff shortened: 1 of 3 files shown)


test/issues/github_304.cpp
```diff
@@ -0,0 +1,21 @@
+// Copyright Jason Rice 2016
```
ldionne: Can we please rename this file  to `test/ext/std/array/issue_304.cpp`? This way, the dependency upon external adapters is explicit.

ricejasonf:
I added your file and renamed the other test. (I could remove it if you think it is redundant.)

Sorry about referencing the issue in the commit message. I forgot that it did that. :sweat_smile:

ldionne:
Thanks a lot! I'll merge this as soon as the CIs are done.

ldionne:
IDK why this was not closed automatically when I merged, but anyway it's merged as ea571a11fc5d3dc6fd2f73a0b1e7e9c0be2fc614.

PR #307, opened 2016-11-07
[string] Add a c_str() method

ldionne:


ricejasonf:
Does adding the storage to the body of `hana::string` affect the compile-time of creating strings?

I might have some time to test this out if you haven't already. ☕️

ldionne:
Good point; I don't know for sure. It would be great if you had time to check it out, I'm at the committee meeting right now.

ricejasonf:
```
  <% (0..n).each do |i| %>

    auto x<%= i %> = string<<%= ([i % 128] * 100).join(', ') %>>{};

  <% end %>
```

It wasn't a perfect test, but it looks like the difference is trivial.

ldionne:
Thanks a lot for looking into it! Here's what I tried: https://gist.github.com/ldionne/bf105d0ee029774677ef55c3e2e409c1.

And the results:
screen shot 2016-11-08 at 06 54 34

So I'll use the second implementation, which seems to be a bit faster.

PR #329, opened 2017-04-06
[index_if] New index_if function

ricejasonf:
- Adds index_if
    - Rewrites detail::index_if to use recursive alias stuff
      optimized for tuple and basic_tuple
    - find_if now uses index_if for Iterables
    - at_key now uses index_if for Sequence
    - Removes duplicate code and unnecessary special case implementations
        - detail::advance_until
        - at_key::advance_until
        - tuple_tag implementation of find_if
    - Uses Foldable instead of Sequence for cases where length is known.
      (find_if had a specialization when Iterable and not Sequence)
    - Adds test.*.auto.index_if for Sequences
    - Adds test support/counter for testing infinite iterables

I have yet to run any benchmarks on these changes.


include/boost/hana/fwd/index_if.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include <boost/hana/core/when.hpp>
+
+
+BOOST_HANA_NAMESPACE_BEGIN
+    //! Finds the value associated to the first key satisfying a predicate.
+    //! @ingroup group-Iterable
+    //!
+    //! Given an `Iterable` structure `xs` and a predicate `pred`,
+    //! `index_if(xs, pred)` returns an `IntegralConstant` index *of* the first
+    //! element whose key satisfies the predicate. If no element satisfies the
```
ldionne: Iterables have no notion of `key`, so just saying "index of the first element that satisfies the predicate" is enough.

include/boost/hana/index_if.hpp
```diff
@@ -0,0 +1,116 @@
...
+Distributed under the Boost Software License, Version 1.0.
+(See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
+ */
+
+#ifndef BOOST_HANA_INDEX_IF_HPP
+#define BOOST_HANA_INDEX_IF_HPP
+
+#include <boost/hana/fwd/index_if.hpp>
+
+#include <boost/hana/fwd/at.hpp>
```
ldionne: Nitpick, but please order the includes alphabetically.

include/boost/hana/index_if.hpp
```diff
@@ -0,0 +1,116 @@
...
+                    ::template f<detail::partial_at<Xs>, Pred>
+        { return {}; }
+    };
+
+    template <typename It>
+    struct index_if_impl<It, when<!Foldable<It>::value>> {
+        template <typename Xs, typename Pred>
+        static constexpr auto apply(Xs const&, Pred const&)
+            -> typename detail::iterate_while<hana::size_c<0>,
+                std::numeric_limits<std::size_t>::max(), false>
```
ldionne: Since `size_t` is unsigned, and unsigned underflow is defined to wrap, I would suggest being bold and just using `static_cast<size_t>(-1)`. I normally wouldn't do this, but this will allow us to get rid of the `<limits>` include.

test/_include/auto/index_if.hpp
```diff
@@ -0,0 +1,73 @@
...
+    BOOST_HANA_CONSTANT_CHECK(hana::equal(
+        hana::index_if(MAKE_TUPLE(ct_eq<0>{}), hana::equal.to(ct_eq<0>{})),
+        hana::just(hana::size_c<0>)
+    ));
+    BOOST_HANA_CONSTANT_CHECK(hana::equal(
+        hana::index_if(MAKE_TUPLE(ct_eq<0>{}), hana::equal.to(ct_eq<42>{})),
+        hana::nothing
+    ));
+    BOOST_HANA_CONSTANT_CHECK(hana::equal(
+        hana::index_if(MAKE_TUPLE(ct_eq<0>{}, ct_eq<1>{}, ct_eq<2>{}), hana::equal.to(ct_eq<1>{})),
```
ldionne: Can you add a test case that goes like `index_if(MAKE_TUPLE(ct_eq<0>, ce_eq<1>, ...), equal.to(ct_eq<0>))`, please?

test/_include/auto/index_if.hpp
```diff
@@ -0,0 +1,73 @@
+// Copyright Louis Dionne 2013-2017
+// Copyright Jason Rice 2017
+// Distributed under the Boost Software License, Version 1.0.
+// (See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_HANA_TEST_AUTO_INDEX_IF_HPP
+#define BOOST_HANA_TEST_AUTO_INDEX_IF_HPP
+
+#include <boost/hana/assert.hpp>
```
ldionne: Order includes please.

test/index_if.cpp
```diff
@@ -0,0 +1,30 @@
...
+namespace hana = boost::hana;
+
+
+int main() {
+    // Tests hana::index_if on an infinite iterable
+    constexpr Counter<> c{};
+    auto pred = [](auto i) {
+        return [=](auto x) {
+        return hana::bool_c<decltype(x)::value == decltype(i)::value>;
+        };};
```
ldionne: Weird indentation.

test/index_if.cpp
```diff
@@ -0,0 +1,30 @@
+// Copyright Louis Dionne 2013-2017
```
ldionne: Fuckin' A! I love tests that go through the trouble of testing infinite sequences, thanks for that.

test/_include/auto/index_if.hpp
ricejasonf: I also forgot an empty tuple.

ricejasonf:
@ldionne I made all changes that you requested and added an example.

include/boost/hana/index_if.hpp
```diff
@@ -0,0 +1,114 @@
...
+    //! @endcond
+
+    namespace detail {
+        template <std::size_t i, std::size_t N, bool Done>
+        struct iterate_while;
+
+        template <std::size_t i, std::size_t N>
+        struct iterate_while<i, N, false> {
+            template <typename F, typename Pred>
+            using f = typename iterate_while<i + 1, N,
```
ldionne: We could be more compile-time efficient here if we hard-coded the `hana::at` call, since we would avoid instantiating `partial_at::operator()` on each iteration. Unless we reuse `iterate_while` somewhere else with a different `F`, I suggest we do that

include/boost/hana/fwd/index_if.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+BOOST_HANA_NAMESPACE_BEGIN
+    //! Finds the value associated to the first key satisfying a predicate.
+    //! @ingroup group-Iterable
+    //!
+    //! Given an `Iterable` structure `xs` and a predicate `pred`,
+    //! `index_if(xs, pred)` returns an `IntegralConstant` index *of* the
+    //! first element that satisfies the predicate. If no element satisfies the
+    //! predicate one of two things may occur. For `Foldable` structures the
+    //! result will be an out-of-bounds `IntegralConstant` index, but for
```
ldionne: The documentation here is incorrect. We return an optional index, not an index that is out of bounds. I'm not sure which one is better, actually. I think I'd go with the optional index, and fix the documentation, but I could easily be convinced otherwise.

ricejasonf:
@ldionne
- I agree that optional is a better interface. I just forgot to change the docs.
- I removed the unnecessary `partial_at` but then realized it was also being used in find_if.hpp,
so I moved to there. oops :sweat_smile: 

Fixed those

ldionne:
Thank you!

Issue #341, opened 2017-04-26
Add set-theoretic functions for hana::map


shreyans800755:
Add intersection and difference interface for hana::map like operation<map_tag, map_tag>.

ldionne:
From Gitter (posting here because it's easier to track discussions, rationales & al):

> For intersection of `hana::map`, if there's a key k in both maps `m1` and `m2` with values respectively `v1` and `v2`. What should the intersection function return? I can see couple of options:
> 1. Return `hana::map` with pair `{k: v1}`
> 2. Return `hana::map` with pair `{k: pair{v1, v2}}`

A third option is to return `hana::map` with pair `{k, v2}`. However, I think we should be consistent with `hana::union_` and return `{k: v1}`. What do you think?

shreyans800755:
Yeah, I think that's better. 2nd option looks like variation of multimap. By the way, is there any plan for including multimap and multiset in hana ?

ldionne:
> By the way, is there any plan for including multimap and multiset in hana?

Sorry for the late reply, I somehow missed that. Right now, there is no plan to have that, but I would welcome a PR that adds them (if they can be reasonably compile-time efficient).

ldionne:
Closed by #350.

PR #350, opened 2017-05-25
Feature/map functions

shreyans800755:
Added intersection and difference functions for map along with docs.
Closes https://github.com/boostorg/hana/issues/341


shreyans800755:
@ldionne I think I should also add symmetric_difference implementation for maps. I'll try to update the PR with its commits soon.

include/boost/hana/fwd/map.hpp
```diff
@@ -280,6 +280,71 @@ BOOST_HANA_NAMESPACE_BEGIN
...
+    //! containing exactly those (key, value) pairs from xs, for which key
+    //! is present in `ys`.
+    //! In other words, the following holds for any object `pair(k, v)`:
+    //! @code
+    //!     pair(k, v) ^in^ intersection(xs, ys) if and only if (k, v) ^in^ xs && k ^in^ ys.keys()
+    //! @endcode
+    //!
+    //!
+    //! @note
+    //! This function is not commutative, i.e. `intersection(xs, ys)` is
```
ldionne: Please reword to

```c++
//! This function is not commutative, i.e. `intersection(xs, ys)` is not
//! necessarily the same as `intersection(ys, xs)`. Indeed, the set of keys
//! in `intersection(xs, ys)` is always the same as the set of keys in
//! `intersection(ys, xs)`, but the value associated to each key may be
//! different. `intersection(xs, ys)` contains values present in `xs`, and
//! `intersection(ys, xs)` contains values present in `ys`.
```

include/boost/hana/fwd/map.hpp
```diff
@@ -280,6 +280,71 @@ BOOST_HANA_NAMESPACE_BEGIN
...
+
+    //! Returns the difference of two maps.
+    //! @relates hana::map
+    //!
+    //! Given two maps `xs` and `ys`, `difference(xs, ys)` is a new map
+    //! containing exactly those (key, value) pairs from xs, for which key
+    //! is not present in `keys(ys)`.
+    //! In other words, the following holds for any object `pair(k, v)`:
+    //! @code
+    //!     pair(k, v) ^in^ difference(xs, ys) if and only if (k, v) ^in^ xs && k ^not in^ ys.keys()
```
ldionne: Same comment for `ys.keys()`.

include/boost/hana/fwd/map.hpp
```diff
@@ -280,6 +280,71 @@ BOOST_HANA_NAMESPACE_BEGIN
...
+    //! @endcode
+    //!
+    //!
+    //! @note
+    //! This function is not commutative, i.e. `difference(xs, ys)` is
```
ldionne: "is not necessarily same" -> "is not necessarily _the_ same". Also, please add an explanation of why that's the case. Something like:

```
//! Indeed, consider the
//! case where `xs` is empty and `ys` isn't. Then, `difference(xs, ys)`
//! is empty but `difference(ys, xs)` is equal to `ys`. For the symmetric
//! version of this operation, see `symmetric_difference`.
```

include/boost/hana/fwd/map.hpp
```diff
@@ -280,6 +280,71 @@ BOOST_HANA_NAMESPACE_BEGIN
...
+    //! not necessarily same as `difference(ys, xs)`.
+    //!
+    //!
+    //! @param xs, ys
+    //! Two maps to find difference.
```
ldionne: Please change to: "Two maps to compute the difference of.". I'm using this formulation for `union_`; it might not be perfect but it'll be consistent.

include/boost/hana/fwd/set.hpp
```diff
@@ -204,9 +204,63 @@ BOOST_HANA_NAMESPACE_BEGIN
...
+    constexpr auto intersection = [](auto&& xs, auto&& ys) {
+        return tag-dispatched;
+    };
+#endif
     //! Equivalent to `to<set_tag>`; provided for convenience.
     //! @relates hana::set
     constexpr auto to_set = to<set_tag>;
 BOOST_HANA_NAMESPACE_END
 
+    //! Returns the difference of two sets.
```
ldionne: Please keep the original documentation for `difference`.

include/boost/hana/map.hpp
```diff
@@ -44,16 +44,20 @@ Distributed under the Boost Software License, Version 1.0.
 #include <boost/hana/fwd/is_subset.hpp>
 #include <boost/hana/fwd/keys.hpp>
 #include <boost/hana/fwd/union.hpp>
+#include <boost/hana/fwd/intersection.hpp>
```
ldionne: Please order includes.

include/boost/hana/map.hpp
```diff
@@ -493,6 +497,59 @@ BOOST_HANA_NAMESPACE_BEGIN
...
+
+    //////////////////////////////////////////////////////////////////////////
+    // difference
+    //////////////////////////////////////////////////////////////////////////
+    template <>
+    struct difference_impl<map_tag> {
+        template <typename Xs, typename Ys>
+        static constexpr auto apply(Xs&& xs, Ys&& ys) {
+            return hana::fold_left(
+                    static_cast<decltype(hana::keys(ys))&&>(hana::keys(ys)),
```
ldionne: Instead, use `hana::keys(static_cast<Ys&&>(ys))`.

include/boost/hana/map.hpp
```diff
@@ -493,6 +497,59 @@ BOOST_HANA_NAMESPACE_BEGIN
...
+    // intersection_
+    //////////////////////////////////////////////////////////////////////////
+    namespace detail {
+        template<typename Ys>
+        struct map_insert_if_contains {
+            Ys const& ys;
+
+            // Second template param will be pair
+            // Get its key and check if it exists, if it does, insert key, value pair.
+            template<typename Result, typename Pair>
```
ldionne: Space between `template` and `<typename`.

include/boost/hana/map.hpp
```diff
@@ -493,6 +497,59 @@ BOOST_HANA_NAMESPACE_BEGIN
...
+            static constexpr auto helper(Result&& result, Pair&& pair, hana::true_) {
+                return hana::insert(static_cast<Result&&>(result), static_cast<Pair&&>(pair));
+            }
+
+            template<typename Result, typename Pair>
```
ldionne: x2

test/map/difference.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+#include <boost/hana/assert.hpp>
+#include <boost/hana/difference.hpp>
+#include <boost/hana/equal.hpp>
+#include <boost/hana/map.hpp>
+
+#include <laws/base.hpp>
+#include <support/minimal_product.hpp>
+namespace hana = boost::hana;
+using hana::test::ct_eq;
```
ldionne: I don't think you need this.

test/map/difference.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+    BOOST_HANA_CONSTANT_CHECK(hana::equal(
+        hana::difference(
+            hana::make_map(p<1, 1>(),
+                           p<2, 2>()),
+            hana::make_map(p<2, 3>(),
+                           p<1, 4>())
+        ),
+        hana::make_map()
+    ));
```
ldionne: Can you add a bigger test case with something like 10 elements in the maps?

test/map/intersection.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+#include <boost/hana/assert.hpp>
+#include <boost/hana/equal.hpp>
+#include <boost/hana/intersection.hpp>
+#include <boost/hana/map.hpp>
+
+#include <laws/base.hpp>
+#include <support/minimal_product.hpp>
+namespace hana = boost::hana;
+using hana::test::ct_eq;
```
ldionne: Not needed.

ricejasonf:
What is the reason for `intersection` not being commutative. Is it a performance thing?

shreyans800755:
@ricejasonf `intersection` is done on the base of `keys` in case of `hana::map`, and not based on key and values, both. Frankly, I only thought this use case, and it didn't really struck into my head about intersecting two maps based on (key, value) pairs.

> Is it a performance thing?

Performance based on asymptotic analysis should be same if want to intersect based on (key, value), although the overhead of comparison of value will be increased.

include/boost/hana/fwd/set.hpp
shreyans800755: If you're talking about lines removed in include/boost/hana/fwd/difference.hpp, then I think it makes sense to remove them, as we have different implementation for set and map. And we have different explanation/docs for that, too.

ricejasonf:
@shreyans800755 Ah, right, the it compares the keys. I wasn't even thinking about that. lol :sweat_smile:

include/boost/hana/fwd/set.hpp
ldionne: I just meant that the lines you removed should be the ones used for `hana::set`.

ldionne:
One of the jobs failed with

> /home/travis/build/boostorg/hana/include/boost/hana/fwd/set.hpp:263: warning: included file example/set/difference.cpp is not found. Check your EXAMPLE_PATH

include/boost/hana/map.hpp
```diff
@@ -51,9 +53,11 @@ Distributed under the Boost Software License, Version 1.0.
 #include <boost/hana/optional.hpp>
 #include <boost/hana/remove_if.hpp>
 #include <boost/hana/second.hpp>
+#include <boost/hana/tuple.hpp>
```
ldionne: Do you need `#include <boost/hana/tuple.hpp>` here?

shreyans800755: Corrected.
I'll push symmetric_difference soon.

example/symmetric_difference.cpp
```diff
@@ -5,10 +5,14 @@
 #include <boost/hana/assert.hpp>
```
ldionne: Just like we did for the other functions, would it be possible to have an `example/set/symmetric_difference.cpp` and an `example/map/symmetric_difference.cpp` file? This would be consistent with the rest of the lib, too.

test/symmetric_difference.cpp
```diff
@@ -0,0 +1,134 @@
+// Copyright Louis Dionne 2013-2017
```
ldionne: Same comment as for examples; it would be nice to have a test for map and a test for sets in different files.

example/symmetric_difference.cpp
shreyans800755: I thought of that, too. I chose this way because we don't have different implementations for set and map.

test/symmetric_difference.cpp
shreyans800755: Same thought as example. So, I wasn't sure.

shreyans800755:
Regarding documentation of symmetric_difference, currently it exists only for set. Should I move it to `include/boost/hana/fwd/set.hpp` and add similar for map in `include/boost/hana/fwd/map.hpp` ?
Or may be just adding a `@relatealso` tag with map in the same file `include/boost/hana/fwd/symmetric_difference.hpp` should do ?

ldionne:
I would duplicate the documentation for set and map. While duplication is generally bad, the rationale here is that the two implementations for set and map are conceptually unrelated, since they don't share a common concept. If we were to create a new concept and move `symmetric_difference` under it (and also probably others), then this is how we would de-duplicate the documentation. Does that make sense?

shreyans800755:
That makes sense. I'll push the changes.

test/set/symmetric_difference.cpp
```diff
@@ -2,8 +2,6 @@
 // Distributed under the Boost Software License, Version 1.0.
 // (See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
 
-#include <boost/hana/assert.hpp>
```
ldionne: Those should stay, no?

shreyans800755: It is already coming from:
test/set/symmetric_difference.cpp:10
test/_include/laws/base.hpp:25
test/_include/support/tracked.hpp:11

And that's why tests ares also passing.

ldionne: Yeah, but you should always include what you use to avoid relying on such dependencies, no?

shreyans800755: Makes sense. Will update soon.

ricejasonf:
:shipit: :+1:

ldionne:
Thanks a lot @shreyans800755, this is a great PR!

PR #359, opened 2017-07-03
[view] cartesian_product

ricejasonf:
- Adds cartesian_product_view_t and cartesian_product_element_view_t
    with specializations for the following:
      - at_impl
      - is_empty_impl
      - length_impl
      - unpack_impl

Note: The name `hana::detail::cartesian_product` is being used because I think we can change namespace `detail` here to `view` to expose these functions. This is important since we can't implement Sequence functions as the view "type" is not technically a Sequence.

We could also then change names like `joined` and `flattened` to simply `view::join` and `view::flatten` so the naming is consistent.

### DIFF
--- a/include/boost/hana/cartesian_product.hpp
+++ b/include/boost/hana/cartesian_product.hpp
@@ -49,7 +49,7 @@ BOOST_HANA_NAMESPACE_BEGIN
         template <std::size_t ...Lengths>
         struct cartesian_product_indices {
             static constexpr std::size_t total_length() {
-                std::size_t lengths[] = {Lengths...};
+                std::size_t lengths[sizeof...(Lengths)] = {Lengths...};
                 std::size_t r = 1;
                 for (std::size_t len: lengths)
                     r *= len;
@@ -59,9 +59,12 @@ BOOST_HANA_NAMESPACE_BEGIN
             static constexpr std::size_t length = total_length();
 
             static constexpr auto indices_of(std::size_t i) {
-                constexpr std::size_t lengths[] = {Lengths...};
                 constexpr std::size_t n = sizeof...(Lengths);
+                constexpr std::size_t lengths[n] = {Lengths...};
                 detail::array<std::size_t, n> result{};
+                if (length == 0) {
+                    return result;
+                }
                 for (std::size_t j = n; j--;) {
                     result[j] = i % lengths[j];
                     i /= lengths[j];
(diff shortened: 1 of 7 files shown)


ricejasonf:
Please let my note read as a suggestion for 
> Consider providing access to the flattened methods & al.

I don't mean to be presumptuous. :smile:

ricejasonf:
@ldionne Have you had a chance to look at this? I think there might be a problem with the fact that `unpack` yields nested views which would make the result of `to_tuple` not be a tuple of tuples.

Perhaps we could add a `to_impl` for Sequences and `view_tag` if you don't think that is a hack.

ricejasonf:
ping! Is this really bad or are you just busy? :stuck_out_tongue:

ldionne:
The latter. Sorry, let me have a look at this right now.

include/boost/hana/cartesian_product.hpp
```diff
@@ -59,31 +59,79 @@ BOOST_HANA_NAMESPACE_BEGIN
...
+        };
+
+        constexpr make_cartesian_product_indices_t make_cartesian_product_indices{};
+
+        template <typename F, std::size_t n>
+        struct make_cartesian_product_element_t {
+            F const& f;
+
+            constexpr auto operator()() const
+            {
```
ldionne: Placement of braces

include/boost/hana/cartesian_product.hpp
```diff
@@ -59,31 +59,79 @@ BOOST_HANA_NAMESPACE_BEGIN
...
+                return f();
+            }
+
+            template <typename X1, typename ...Xs>
+            constexpr auto operator()(X1&& x1, Xs&& ...xs) const
+            {
```
ldionne: x2

include/boost/hana/cartesian_product.hpp
```diff
@@ -59,31 +59,79 @@ BOOST_HANA_NAMESPACE_BEGIN
...
-                    std::make_index_sequence<sizeof...(Xs)>{}, xs...
+            create_product(std::index_sequence<n...>, F&& f, Xs&& ...xs) {
+                return F{}(product_element<n>(
+                    std::make_index_sequence<sizeof...(Xs)>{},
+                    static_cast<F&&>(f), xs...
                 )...);
             }
         };
+
+        struct make_cartesian_product_indices_helper_t {
```
ldionne: Okay, that's cool.

include/boost/hana/view.hpp
```diff
@@ -59,6 +60,43 @@ BOOST_HANA_NAMESPACE_BEGIN
...
+            constexpr cartesian_product_view_t<Sequences> operator()(Sequences& s) const {
+                return {s};
+            }
+        };
+
+        constexpr make_cartesian_product_view_t cartesian_product{};
+
+        template <typename Sequences>
+        struct is_view<cartesian_product_view_t<Sequences>> {
+            static constexpr bool value = false;
```
ldionne: And here too?

include/boost/hana/view.hpp
```diff
@@ -59,6 +60,43 @@ BOOST_HANA_NAMESPACE_BEGIN
...
+        //////////////////////////////////////////////////////////////////////
+        template <typename Sequences, std::size_t i>
+        struct cartesian_product_element_view_t {
+            detail::view_storage<Sequences> sequences_;
+            using hana_tag = view_tag;
+        };
+
+        template <typename Sequences, std::size_t i>
+        struct is_view<cartesian_product_element_view_t<Sequences, i>> {
+            static constexpr bool value = false;
```
ldionne: Wait, don't you mean `true` here?

include/boost/hana/view.hpp
```diff
@@ -279,6 +344,22 @@ BOOST_HANA_NAMESPACE_BEGIN
...
+            using Indices = decltype(detail::make_cartesian_product_indices(view.sequences_));
+            return hana::at_c<Indices::indices_of(i)[N::value]>(
+                      hana::at_c<N::value>(view.sequences_));
+        }
+
+        // cartesian_product_view
+        template <typename Sequences, typename N>
+        static constexpr decltype(auto)
+        apply(detail::cartesian_product_view_t<Sequences> view, N const&) {
+          return detail::cartesian_product_element_view_t<Sequences, N::value>{view.sequences_};
```
ldionne: Before this point I wondered why you needed this `cartesian_product_element_view_t` type, but now I see.

test/view/cartesian_product/at.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+#include <laws/base.hpp>
+#include <support/seq.hpp>
+namespace hana = boost::hana;
+using hana::test::ct_eq;
+
+
+int main() {
+    auto container = ::seq;
+
```
ldionne: This is missing a unit test for an empty sequence, and for a non-empty sequence containing empty sequences (first, second, third, etc..). Please add a few test cases like that. It's not so much because I doubt the correctness of your implementation, but these sort of test cases are very good at making strange bugs surface (e.g. the unary copy-constructors for tuple surfaces several SFINAE-related bugs).

test/view/cartesian_product/modifications.cpp
```diff
@@ -0,0 +1,39 @@
...
+    auto container = ::seq;
+
+    auto storage = container(container(0),
+                             container(1),
+                             container(2, 3),
+                             container(4, 5, 6));
+
+    auto view = hana::detail::cartesian_product(storage);
+
+    hana::at_c<0>(hana::at_c<0>(view)) = 90;
```
ldionne: Oh, it's neat that you can do that. Also kind of scary, but it makes sense.

ricejasonf:
The point of this was to facilitate `hana::ap`.
```cpp
    template <>
    struct ap_impl<hana::view_tag> {
        template <typename F, typename X>
        static constexpr auto apply(F&& f, X&& x) {
            // TODO: Implement cleverly; we most likely need a cartesian_product
            //       view or something like that.
            return hana::ap(hana::to_tuple(f), hana::to_tuple(x));
        }
};
```

ldionne:
Ahhhhhhhhhhhhhh!!! Sorry! No more concern, only those things I left comments on. Also, I think the `to_tuple` issue is not a big deal. What would you expect?

ricejasonf:
`to_tuple` would yield a `tuple` of views which makes sense but could be a potential pitfall I guess.

I'm not worried about if you think it will be fine.

ldionne:
I don't think it's a big deal. Also, I just don't see a good solution.

test/view/cartesian_product/at.cpp
ricejasonf: I'm adding these to the `unpack` tests.

ricejasonf:
Fixups

- Fixes the formatting issues
- Adds tests, and they did indeed flush out a zero-length array bug.
- Changes the name `detail::cartesian_product` to `detail::cartesian_product_view`

PR #361, opened 2017-07-08
Make sure we support compiling with C++17

ldionne:
That should work out of the box, but (at least) Clang has some bugs that break us when using `-std=c++1z`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,6 +97,14 @@ matrix:
     ##########################################################################
     # Build with variations in the configuration
     ##########################################################################
+    # With C++17 instead of C++14, on Clang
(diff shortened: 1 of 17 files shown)


ricejasonf:
Is this complete? Github says checks failed but travis says the build is passing.

I remember `make check` failing when I tried c++1z because of a bunch of additional warnings that caused it to fail. I don't see anything in the cmake stuff that addresses that.

(or maybe I was just doing something else wrong but it was unrelated stuff... idk)

In any case I am very excited about this :+1: :+1: :+1:

ldionne:
I just kicked off the Travis build now.

Issue #357, opened 2017-07-01
Concepts for set-theoretic functions


shreyans800755:
At the moment, there are no concepts written for set-theoretic functions, due to which minor errors lengthens up to multiple pages unnecessarily.

shreyans800755:
Working on issue, will create PR as soon as considerable work is done.

ldionne:
Awesome! Thanks a lot! Feel free to submit a PR with whatever you've got so that we can iterate on the wording of the concepts, which I expect will take a few iterations to get right.

shreyans800755:
@ldionne Sure. Just want to clarify: Do we need different commits for different set-theoretic-functions or single ?

ldionne:
Not sure I understand. What I refer to when I say concepts is adding a new concept like `Monad` or `Foldable`, but suitable for set-theoretic operations. That should not require changing the existing functions except in a very mechanical way. Either one or many commits is fine, I don't have a preference so long as the PR is convenient to review.

shreyans800755:
@ldionne Are we looking for different concepts for different set-theoretic functions, or just a single one for all ?

ldionne:
Ideally a single one for all. Just like `Functor` has `transform`, `replace_if`, `replace`, etc.., I am looking for some `SetTheoretic` concept (name TBD) that has `difference`, `union`, `intersection`, etc...

PR #364, opened 2017-08-20
[SetTheoretic] Add SetTheoretic concept

shreyans800755:
It adds SetTheoretic concept for functions-
union, intersection, difference, symmetric_difference.

Closes https://github.com/boostorg/hana/issues/357


shreyans800755:
@ldionne Can you please review this ?

example/misc/infinite_set.cpp
```diff
@@ -49,6 +50,11 @@ constexpr auto doubleton(X x, Y y) {
 
 namespace boost { namespace hana {
     template <>
+    struct SetTheoretic<infinite_set_tag> {
+        static constexpr bool value = true;
+    };
```
ricejasonf: If the point of the `SetTheoretic` concept is to guarantee that all associated functions are implemented, then being able to just opt in to use `union` in this case is probably incorrect.

For this example, I would suggest either implementing the remaining functions or just make a custom `union` function. (not sure which is preferable)

shreyans800755: I was thinking it the other way around. If you want to use any of the set theoretic functions (at least one of 4 available at the moment), then it should have a concept `SetTheoretic`. Container doesn't need to implement all 4 operations IMO.
Essentially it means that if you want to use any of these 4 functions, container must be `SetTheoretic`.

ricejasonf: Your documentation specifies a minimal complete definition including all four functions. I think this is what makes the concept useful, the guarantee that those functions work.

include/boost/hana/fwd/concept/set_theoretic.hpp
```diff
@@ -0,0 +1,37 @@
...
+BOOST_HANA_NAMESPACE_BEGIN
+    //! @ingroup group-concepts
+    //! @defgroup group-SetTheoretic SetTheoretic
+    //! The `SetTheoretic` concept represents data structures supporting
+    //! set-theoretic functions like union, instersection, difference,
+    //! and symmetric_difference.
+    //!
+    //! Minimal complete definition
+    //! ---------------------------
+    //! `union_`, `intersection`, `difference` and `symmetric_difference`
```
ldionne: `symmetric_difference` can be derived from `union` and `difference`, see [this](https://en.wikipedia.org/wiki/Symmetric_difference#Properties). Hence, I don't think it should be part of the minimal complete definition.

include/boost/hana/fwd/concept/set_theoretic.hpp
```diff
@@ -0,0 +1,37 @@
...
+    //!
+    //! Concrete models
+    //! ---------------
+    //! `hana::set`, `hana::map`
+    //!
+    //!
+    //! [1]: https://github.com/boostorg/hana/issues/357
```
ldionne: Why that link?

include/boost/hana/fwd/concept/set_theoretic.hpp
```diff
@@ -0,0 +1,37 @@
...
+#ifndef BOOST_HANA_FWD_CONCEPT_SET_THEORETIC_HPP
+#define BOOST_HANA_FWD_CONCEPT_SET_THEORETIC_HPP
+
+#include <boost/hana/config.hpp>
+
+
+BOOST_HANA_NAMESPACE_BEGIN
+    //! @ingroup group-concepts
+    //! @defgroup group-SetTheoretic SetTheoretic
+    //! The `SetTheoretic` concept represents data structures supporting
```
ldionne: Defining the concept as "things that have the methods that the concept provides" is kind of circular, and not very useful IMO. Also, you're not defining laws for the concept. I'd go for a brief like this instead:

> The `SetTheoretic` concept represents data structures that support an [algebra of sets][1].

Then, I'd make some comments about what the concept is at a high level, and why it's useful. I'd also give the laws that the concept must follow, e.g. probably the ones on the Wikipedia page.

Other comments:
- One thing that must be thought about is the relationship between this concept and `Searchable`. Indeed, `Searchable` provides `is_subset`; it seems clear that there's an interesting mathematical link between both, and it should be explored, explained and some laws should be surfaced for data structures that are both `Searchable` and `SetTheoretic`.
- You should also think about automatically-provided models and document them, if you can find some. What I mean here is that you should try to see whether satisfying some other concept(s) makes it possible to satisfy `SetTheoretic` automatically. When this is the case, Hana usually provides such a definition and documents it (e.g. `Foldable` for any `Iterable`).
- The methods should be documented like normal methods that are tied to a concept, not individually for `map` and `set`.

I know this may seem like a lot, but this is what a concept in Hana is; it's not only a syntactical construction, it's also a tool to reason mathematically about programs, which requires a little more work to define than "loose" concepts that don't have clear semantics, if I may put it that way.

If there are things where you really don't know what to do, let me know and I'll try to make some time for researching the topic to try and provide some guidance.

[1]: https://en.wikipedia.org/wiki/Algebra_of_sets

include/boost/hana/intersection.hpp
```diff
@@ -20,10 +21,20 @@ BOOST_HANA_NAMESPACE_BEGIN
...
+            hana::SetTheoretic<TX>::value &&
+            std::is_same<TX, TY>::value
         );
+        
+    #ifndef BOOST_HANA_CONFIG_DISABLE_CONCEPT_CHECKS
+        static_assert(hana::SetTheoretic<TX>::value,
+        "hana::intersection(xs, ys) requires 'xs' to be SetTheoretic");
+
+        static_assert(std::is_same<TX, TY>::value,
+        "hana::intersection(xs, ys) requires 'xs' and 'ys' to be SetTheoretic");
```
ldionne: The `static_assert` message is wrong.

example/misc/infinite_set.cpp
ldionne: The usual way of doing things in Hana is that if you satisfy the minimal complete definition, then you automatically model the concept, no need to specialize it. I think this is what we should do here. The only concept that requires specialization is `Sequence`, and that's because there's no additional syntax added on top of existing concepts, only semantics. But this is not the case here, so I'd like the model to be deduced automatically.

include/boost/hana/map.hpp
```diff
@@ -207,6 +208,12 @@ BOOST_HANA_NAMESPACE_BEGIN
             using type = detail::map_impl<HashTable, Storage>;
         };
     }
+    
+    template <>
+    struct SetTheoretic<map_tag> {
+        static constexpr bool value = true;
```
ldionne: This should not be necessary.

include/boost/hana/set.hpp
```diff
@@ -77,6 +78,11 @@ BOOST_HANA_NAMESPACE_BEGIN
         constexpr set(set&& other) = default;
     };
     //! @endcond
+    
+    template <>
+    struct SetTheoretic<set_tag> {
+        static constexpr bool value = true;
```
ldionne: This should not be necessary.

include/boost/hana/symmetric_difference.hpp
```diff
@@ -22,10 +23,20 @@ BOOST_HANA_NAMESPACE_BEGIN
...
+            hana::SetTheoretic<TX>::value &&
+            std::is_same<TX, TY>::value
         );
+        
+    #ifndef BOOST_HANA_CONFIG_DISABLE_CONCEPT_CHECKS
+        static_assert(hana::SetTheoretic<TX>::value,
+        "hana::symmetric_difference(xs, ys) requires 'xs' to be SetTheoretic");
+
+        static_assert(std::is_same<TX, TY>::value,
+        "hana::symmetric_difference(xs, ys) requires 'xs' and 'ys' to be SetTheoretic");
```
ldionne: The `static_assert` message is wrong.

include/boost/hana/union.hpp
```diff
@@ -20,11 +21,21 @@ BOOST_HANA_NAMESPACE_BEGIN
...
+            hana::SetTheoretic<TX>::value &&
+            std::is_same<TX, TY>::value
         );
 
+    #ifndef BOOST_HANA_CONFIG_DISABLE_CONCEPT_CHECKS
+        static_assert(hana::SetTheoretic<TX>::value,
+        "hana::union_(xs, ys) requires 'xs' to be SetTheoretic");
+
+        static_assert(std::is_same<TX, TY>::value,
+        "hana::union_(xs, ys) requires 'xs' and 'ys' to be SetTheoretic");
```
ldionne: The `static_assert` message is wrong.

include/boost/hana/difference.hpp
```diff
@@ -21,11 +22,21 @@ BOOST_HANA_NAMESPACE_BEGIN
...
+            hana::SetTheoretic<TX>::value &&
+            std::is_same<TX, TY>::value
         );
 
+    #ifndef BOOST_HANA_CONFIG_DISABLE_CONCEPT_CHECKS
+        static_assert(hana::SetTheoretic<TX>::value,
+        "hana::difference(xs, ys) requires 'xs' to be SetTheoretic");
+
+        static_assert(std::is_same<TX, TY>::value,
+        "hana::difference(xs, ys) requires 'xs' and 'ys' to be SetTheoretic");
```
ldionne: The static_assert message is wrong; it should say "requires 'xs' and 'ys' to be of the same data type", or equivalent.

include/boost/hana/fwd/concept/set_theoretic.hpp
ricejasonf: Would it make sense for `SetTheoretic` to have its own accessors? Something like `get<T>(set)` == `at_key(set, T{})` would be very useful for `set` and it would be a way of making `SetTheoretic` provide an implementation for `Searchable`.

It might even simplify implementation of these functions assuming "keys" are strictly looked up at compile-time. Hmm.. that might also require something like the `keys` function that `map` has.

Also perhaps it should be just `hana::Set`.

ricejasonf: The TLDR is that they would only have to implement those accessors and not the union, intersection,... algorithms directly

DarthRubik: From what I see it does not necessarily make sense to have `SetTheoretic` be related to `Searchable` if the goal is to create a generic thing that supports the algebra of sets.

You can make plenty of things that support the algebra of sets that it does not make sense to be `Searchable`.  

For example: a string of bits:
    
    1010 union 0001 => 1011
    1010 intersection 1000 => 1000
    1010 difference 1000 => 0010
    1010 symmetric_difference 1001 => 0011

Anyways......thats just the way I see it.......
@ldionne @ricejasonf @shreyans800755

ricejasonf: @DarthRubik I don't believe we are trying to couple `Set` with `Searchable`, but types that satisfy `Searchable` can be made to implement the needed functions for `Set`. For your string of bits you would probably implement those functions directly.

shreyans800755:
Resuming work after ages. Would like complete it this time.

example/misc/infinite_set.cpp
shreyans800755: Makes sense. Removing it.

include/boost/hana/fwd/concept/set_theoretic.hpp
shreyans800755: Makes sense. I think it should automatically kick in while calling difference first time if condition is not satisfied

include/boost/hana/fwd/concept/set_theoretic.hpp
shreyans800755: removed

include/boost/hana/intersection.hpp
shreyans800755: fixed

include/boost/hana/map.hpp
shreyans800755: fixed

include/boost/hana/symmetric_difference.hpp
shreyans800755: fixed

include/boost/hana/difference.hpp
shreyans800755: fixed

example/misc/infinite_set.cpp
```diff
@@ -56,6 +62,22 @@ namespace boost { namespace hana {
         }
     };
 
+    template <>
+    struct difference_impl<infinite_set_tag> {
```
shreyans800755: @ldionne 
This actually doesn't seem good to me logically. Although infinite set had union function implemented, we can't call it, because apparantly, union requires the class to be SetTheoretic meaning it requires minimal definition - all 3 functions union, intersection & difference. But, mathematically, it seems accurate though.
Any thoughts ?

shreyans800755:
Also, mathematically, this doesn't seem accurate. Because, ideally intersection must be commutative, but here only set::intersection is commutative, but map::intersection is not.
@ldionne I'm not sure if this was intended, but your input would be appreciated.

PR #377, opened 2018-01-01
Add support for GCC 5

### LINKED ISSUE #368: Is it possible to support gcc 5.5?
gcc 5.5 was released!! Is it possible to support gcc 5.5?

ldionne:
This PR is a work in progress that attempts to add support for GCC 5.5, as suggested in #368. It may never make it, depending on the amount of work necessary to support GCC 5.5.

Fixes #368


ldionne:
Note: Travis containers are still on GCC 5.4.x: https://travis-ci.org/boostorg/hana/jobs/323891150#L302

ldionne:
GCC 5 is kinda old, I'm closing this because I don't think it's worth the effort.

PR #378, opened 2018-01-01
Add tests pre-check for B2

Kojoley:


ldionne:
This seems to have broken our Travis build: https://travis-ci.org/boostorg/hana/jobs/323956679#L4190. Any ideas?

Kojoley:
Oops.
I have checked `libs/config/checks/config.jam` of `boost_1_65_1.tar.gz` the travis is downloading and it is here.
I am wondering if `import config : requires ;` is supported only by the latest b2, while older versions only worked with `import ../../config/checks/config : requires ;`?

ldionne:
I think this is because we don't actually build Hana as a subdirectory of the Boost tree. I'm trying to tweak the Travis [here](https://github.com/boostorg/hana/commit/b156c68e8f4d96d1e9cf4d10e27485348242f727).

ldionne:
FYI, this has been fixed by https://github.com/boostorg/hana/pull/385

PR #391, opened 2018-03-05
Make some containers final and add notes about assumptions about representation

### LINKED ISSUE #388: inherit from empty tuple
A class inherited from an empty tuple can not be passed to boost::hana::unpack() (or any other hana method that uses unpack).

```
    template<typename... T>
    struct test : hana::tuple<T...> { };

    hana::for_each(test<> { }, [](auto){
        cout << "fuu" << endl;
    });
```

As soon as you have at least one template parameter, everything works as expected.

```
    hana::for_each(test<int> { }, [](auto){
        cout << "fuu" << endl;
    });
```

Possible solution for that issue:

```
    template <>
    struct unpack_impl<tuple_tag> {

        template<typename Xs, typename = void>
        struct helper
        {
            template <typename F>
            static constexpr decltype(auto) apply(Xs&& xs, F&& f)
                { return hana::unpack(static_cast<Xs&&>(xs).storage_, static_cast<F&&>(f)); }
        };

        template<typename Xs>
        struct helper<Xs,   std::enable_if_t<
                                std::is_base_of<
                                    tuple<>,
                                    std::remove_cv_t<
                                        std::remove_reference_t<Xs>
                                    >
                                >::value
                            >>
        {
            template <typename F>
            static constexpr decltype(auto) apply(Xs&&, F&& f)
                { return static_cast<F&&>(f)(); }
        };

        template <typename Xs, typename F>
        static constexpr decltype(auto) apply(Xs&& xs, F&& f) {
            return helper<Xs>::apply(static_cast<Xs&&>(xs), static_cast<F&&>(f));
        }
    };
```

ldionne:
Also, add tests to make sure that an empty pair can be EBO'd. This one is very important because a typical use case is to create a tuple of pairs of empty types (e.g. in `hana::map`), and we expect this to be empty.

Fixes #388

### DIFF
--- a/doc/tutorial.hpp
+++ b/doc/tutorial.hpp
@@ -2024,7 +2024,10 @@ a container unspecified; they are explained in the
 [rationales](@ref tutorial-rationales-container_representation).
 When the representation of a container is implementation-defined, one must
 be careful not to make any assumptions about it, unless those assumption
-are explicitly allowed in the documentation of the container.
+are explicitly allowed in the documentation of the container. For example,
+assuming that one can safely inherit from a container or that the elements
+in the container are stored in the same order as specified in its template
+argument list is generally not safe.
 
 
 @subsubsection tutorial-containers-types-overloading Overloading on container types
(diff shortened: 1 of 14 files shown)


ricejasonf:
Perhaps `hana::map<???>` and others should be in the `detail` namespace.

ldionne:
> Perhaps hana::map<???> and others should be in the detail namespace.

No, because `hana::map<...>` can be used to specify the type of a map. The type exists, it's just that what the type is is unspecified.

PR #392, opened 2018-04-06
[lift] Defer concept check until invocation

ricejasonf:
Here is my use case to justify this:

```cpp
#include <boost/hana/fwd/lift.hpp>

namespace full_duplex {
    struct promise_tag { };

    struct promise_fn {
        template <typename AsyncFn>
        constexpr auto operator()(AsyncFn&&) const;
    };

    constexpr promise_fn promise{};

    constexpr boost::hana::lift_t<promise_tag> promise_lift;
    // ^ requires promise_tag to be an Applicative before there is a chance to implement
}
```
BTW
I'm making a clean break and implementing my promise stuff as a proper `hana::Monad`.
Shameless plug: [Generalized Full Duplex Messaging](http://sched.co/EC73)

### DIFF
--- a/include/boost/hana/fwd/lift.hpp
+++ b/include/boost/hana/fwd/lift.hpp
@@ -49,7 +49,10 @@ BOOST_HANA_NAMESPACE_BEGIN
     struct lift_impl : lift_impl<A, when<true>> { };
 
     template <typename A>
-    struct lift_t;
+    struct lift_t {
+        template <typename X>
+        constexpr auto operator()(X&& x) const;
+    };
 
     template <typename A>
     constexpr lift_t<A> lift{};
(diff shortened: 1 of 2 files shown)


ldionne:
Your change makes Doxygen [emit a warning](https://travis-ci.org/boostorg/hana/jobs/363325746#L1648); please use `//! @cond` and `// @endcond` around the definition of `operator()` -- I think that should make it go away.

ricejasonf:
Fixed the warning
```
root@873199c26702:/opt/build# make doc
Scanning dependencies of target doc
Generating API documentation with Doxygen
Built target doc
root@873199c26702:/opt/build#
```

ldionne:
Thanks for the PR!

PR #393, opened 2018-04-12
More deferring concept checks

ricejasonf:
I encountered this chicken/egg thing again with `sum<Tag>` so I hit them all to be consistent.

:sweat_smile:

### DIFF
--- a/include/boost/hana/sum.hpp
+++ b/include/boost/hana/sum.hpp
@@ -23,28 +23,26 @@ Distributed under the Boost Software License, Version 1.0.
 
 
 BOOST_HANA_NAMESPACE_BEGIN
+    //! @cond
     template <typename M>
-    struct sum_t {
+    template <typename Xs>
+    constexpr decltype(auto) sum_t<M>::operator()(Xs&& xs) const {
+        using S = typename hana::tag_of<Xs>::type;
+        using Sum = BOOST_HANA_DISPATCH_IF(sum_impl<S>,
+            hana::Foldable<S>::value
+        );
+
     #ifndef BOOST_HANA_CONFIG_DISABLE_CONCEPT_CHECKS
         static_assert(hana::Monoid<M>::value,
         "hana::sum<M> requires 'M' to be a Monoid");
+
+        static_assert(hana::Foldable<S>::value,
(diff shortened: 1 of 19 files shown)


include/boost/hana/sum.hpp
```diff
@@ -23,28 +23,28 @@ Distributed under the Boost Software License, Version 1.0.
...
-            using S = typename hana::tag_of<Xs>::type;
-            using Sum = BOOST_HANA_DISPATCH_IF(sum_impl<S>,
-                hana::Foldable<S>::value
-            );
+        using S = typename hana::tag_of<Xs>::type;
+        using Sum = BOOST_HANA_DISPATCH_IF(sum_impl<S>,
+            hana::Foldable<S>::value
+        );
 
-        #ifndef BOOST_HANA_CONFIG_DISABLE_CONCEPT_CHECKS
```
ldionne: Could you please group the two concept checks inside the same `#ifndef` please? Here and elsewhere.

ricejasonf:
Err.. I did two force pushes so maybe the checks for bb412ea can be stopped.

ricejasonf:
It looks like the documentation test failed, but this time it wasn't my fault :stuck_out_tongue: 
> The program 'doxygen' is currently not installed. To run 'doxygen' please ask your administrator to install the package 'doxygen'

FWIW I ran `make doc` on my system and it doens't have any warnings.
```
root@256fd253009f:/opt/build# make doc
Scanning dependencies of target doc
Generating API documentation with Doxygen
Built target doc
root@256fd253009f:/opt/build#
```

include/boost/hana/sum.hpp
ldionne: Ping. I'm good to merge this once this is done. I'd like to avoid having multiple `#ifdefs` for `BOOST_HANA_CONFIG_DISABLE_CONCEPT_CHECKS` in the same function when that makes sense.

ricejasonf: I had pushed for this, but apparently I missed the second overloads for `monadic_fold_left/right`.
I rechecked them all and pushed again. Should be fixed now. :sweat_smile:

ldionne:
Thank you!

PR #394, opened 2018-04-30
[tuple] remove superfluous conversion in example

ricejasonf:
The documentation states that the result of `tuple_c` is functionally equivalent to the call to `make_tuple` so the example should not require a conversion. Correct me if I am wrong.

I've seen confusion about this before, and here is the question that prompted this:
https://stackoverflow.com/questions/50101953/why-is-the-type-of-boosthanatuple-c-implementation-defined

Thanks :smile:


ldionne:
You are right. Not sure why there's a conversion in the first place.

PR #404, opened 2018-05-30
[experimental] Printable support for Visual C++

ricejasonf:
- Conditionally strips out preceding `struct` keyword since
    Microsoft's compiler includes that in the outputs of type_info::name

NOTE: I was not able to test this on Visual C++ as it is closed source and the version is likely not released.


ricejasonf:
@ldionne 

I implemented your suggestion to use a regex in the tests.

The printed output would look like: `metafunction<*implementation-defined*>`.

The only part that is cleaned up is the outer `template`, `metafunction`, `metafunction_class` part just like before. The types rendered within are whatever the compiler gives us.

PR #413, opened 2018-07-01
Making overload_t's copy constructor have priority

DarthRubik:
So currently the copy constructor for the overload_t does not take
precendent over the other templated constructor.  This means a simple
function like the following fails:

	void foo()
	{
		auto a = hana::overload(f1,f2);
		auto b(a); //Error
	}

To fix this I added a enable_if to turn off the constructor when not
applicable.........

Note: This (allegedly) fixes issue 412

### DIFF
--- a/include/boost/hana/functional/overload.hpp
+++ b/include/boost/hana/functional/overload.hpp
@@ -12,6 +12,7 @@ Distributed under the Boost Software License, Version 1.0.
 
 #include <boost/hana/config.hpp>
 #include <boost/hana/detail/decay.hpp>
+#include <type_traits>
 
 
 BOOST_HANA_NAMESPACE_BEGIN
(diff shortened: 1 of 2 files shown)


DarthRubik:
I am not quite sure why the automated builds failed......(I don't have much experience with appveyor or travis-ci).....when I did a local make check nothing failed........does someone have a clue as to what is going on here?

ricejasonf:
It looks like it is detecting trailing white-space. I remember the error message for that being a little better.

The Appveyor build failure looks like a config issue with MSVC.

test/issues/github_412.cpp
```diff
@@ -0,0 +1,17 @@
+#include "boost/hana.hpp"
```
ldionne: We use `<>` to include Hana's headers in the tests.  Also, we try to include the most specific header. `#include <boost/hana/functional/overload.hpp>`

include/boost/hana/functional/overload.hpp
```diff
@@ -47,7 +48,10 @@ BOOST_HANA_NAMESPACE_BEGIN
         using overload_t<F>::type::operator();
         using overload_t<G...>::type::operator();
 
-        template <typename F_, typename ...G_>
+        template <typename F_, typename ...G_,
+                  typename = std::enable_if_t<
+                     !std::is_same<overload_t,std::decay_t<F_>>::value
```
ldionne: Nitpick: Please add a space between `overload_t` and `std::decay_t<F_>`.

Issue #410, opened 2018-06-08
hana::experimental::type_name - MSVC support?


d223chen:
With the recent fork provided by Microsoft for Boost Hana / MSVC compiler compatibility, Hana now builds with the MSVC compiler. However, the experimental method hana::experimental::type_name has a hardcoded switch in its helper "type_name_impl2()" that only supports Clang:

        // Note: We substract the null terminator from the string sizes below.
        template <typename T>
        constexpr cstring type_name_impl2() {

        #if defined(__clang__)
            constexpr char const* pretty_function = __PRETTY_FUNCTION__;
            constexpr std::size_t total_size = sizeof(__PRETTY_FUNCTION__) - 1;
            constexpr std::size_t prefix_size = sizeof("boost::hana::experimental::detail::cstring boost::hana::experimental::detail::type_name_impl2() [T = ") - 1;
            constexpr std::size_t suffix_size = sizeof("]") - 1;
	#else
            #error "No support for this compiler."
        #endif

There exist pseudo-equivalent MSVC predefined-macros ([see this doc](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros)) to __PRETTY_FUNCTION__ in : __FUNCSIG__, __FUNCDNAME__, __func__, __FUNCTION__. Also see [this](https://stackoverflow.com/questions/4434282/visual-c-equivalent-of-file-line-and-pretty-function/4434362#comment88540256_4434362) and [this](https://stackoverflow.com/questions/4384765/whats-the-difference-between-pretty-function-function-func). It is also worth looking at [BOOST_CURRENT_FUNCTION](https://www.boost.org/doc/libs/1_42_0/libs/utility/current_function.html). These could possibly make type_name compatible with MSVC by replacing the __PRETTY_FUNCTION__ macro.

I tried adding this switch in the type_name.hpp header file, which allows MSVC to hit the switch and use type_name:

		#elif defined(_MSC_VER)
			constexpr char const* pretty_function = __FUNCSIG__;
			constexpr std::size_t total_size = sizeof(__FUNCSIG__) - 1;
			constexpr std::size_t prefix_size = sizeof("boost::hana::experimental::detail::cstring boost::hana::experimental::detail::type_name_impl2() [T = ") - 1;
			constexpr std::size_t suffix_size = sizeof("]") - 1;

type_name then yields typename strings like this:
int -> "e_impl2<int>(void"
"myStruct" user-defined struct -> "e_impl2<struct myStruct>(void"
[...]

Issue #265, opened 2016-03-28
Consider adding a way to get type names as compile-time strings


ldionne:
As done in [ctti](https://github.com/Manu343726/ctti).

Manu343726:
Let me note the two main issues with CTTI type IDs above:

 - **Fixed size name storage**: Since we wanted to have an **unique** `constexpr` class to store both the name and its hash, the only way to store names homogeneously was writing a fixed-length `constexpr` string class. This has the caveat that storing `ctti::type_id_t`s variables may add a lot of overhead (We have a macro to control the length of the string OTOH). Since I was using ctti mainly to store type IDs and not playing with the type name (As type tag for a variant, IDs for a reflection engine, etc), I wrote a `ctti::unnamed::type_id_t` counterpart that holds the hash only. That's enough to use IDs as keys in hash tables and the like, unless you want the name of the type as part of debugging of course.   
I'm sure this is not a problem in Hana since it's metaprogramming related, so working with heterogeneous types is the whole point of the lib. If I were you, I would ignore all the `constexpr` string stuff and store the strings as typelists (Less compiler support headaches to care about).

 - **ICEs**: As the README of the library says, we support GCC, Clang, and Visual Studio (Linux, MinGW, Cygwin, etc). This looks great for a readme, but from type to time GCC raises ICEs on some use cases. I haven't confirmed this (A good test suite is something the library is missing... (), but I can tell you some patterns that *could* result in an ICE.
First one: **Storing an ID in a class as a static constant**:

   ``` cpp
class MyClass
{
    static constexpr ctti::type_id_t id = ctti::type_id<MyClass>();
};
   ```
 Hopefully the workaround here is simple: Don't store the ID, use a function.

 Today I have found a second one when working on https://github.com/GueimUCM/siplasplas/pull/19: **Using IDs as template parameters** (See the PR comments). Again, this is `constexpr`related, so if you follow with typelists this may not affect you.

Hope it helps.

apolukhin:
Probably it could be more useful to just use the implementation from Boost.TypeIndex https://github.com/apolukhin/type_index/tree/constexpr14. constexpr14 branch will be merged to develop right after the 1.61 release.

Then all you need to do is to include this header boost/type_index/ctti_type_index.hpp

Manu343726:
@ldionne If boost will support a similar feature, I agree with Antony, just pick it from Boost.TypeIndex instead of reinventing the wheel.

[...]

PR #432, opened 2018-12-20
type_name : add support for MSVC (compile time) and GCC (run time)

pthom:
This is my first PR ever to Boost.Hana, so I hope I did not make too many
obvious mistakes. Please bare with me if I did.

This PR extends the support of the experimental type_name feature to MSVC
(compile time) and to GCC (run time only).

All tests pass, see : https://travis-ci.org/pthom/hana/builds/470639623
(The only failing test is due to a missing doxygen download.)

One of the design choice of this PR was to make the inner work related to
`__PRETTY_FUNCTION__` independent from Boost. 
There is an accompanying repo where it can be tested independently, 
in the hope that compiler implementers may take a look at it 
(see https://github.com/pthom/pretty_function_sandbox and 
https://travis-ci.org/pthom/pretty_function_sandbox/builds/470722255)

Thanks.

List of modified files:

```
include/boost/hana/experimental/
├── detail
│   ├── type_name_compiler_capabilities.hpp
│   ├── type_name_pretty_function.hpp
│   └── type_name_stringliteral.hpp
├── type_name.hpp

test/
├── CMakeLists.txt
    test/experimental/
    ├── type_name.cpp
    └── type_name_stringliteral_test.cpp
```

* include/boost/hana/experimental/detail

    The design choices for the files in this folder is that they should
    run without any dependency (not even a dependency on Boost.Hana).
    The main API for them is inside type_name_pretty_function.hpp.

    These files do not use the `BOOST_HANA_NAMESPACE_BEGIN` macro, in
    order not to depend on Boost.Hana. 

    `test/experimental/type_name_stringliteral_test.cpp` is a
    standalone test for these files.

    The intent is to be able to provide them as testing tools for
    compiler implementers, notably GCC which does not support
    constexpr string literals.

    This can be revised if needed of course. 

    The namespace in these files was renamed to `type_name_details::`

    * type_name_pretty_function.hpp

        * Provides
        experimental::type_name_details::type_name_impl_stringliteral()

        * Does this by defining a different prefix/suffix per compiler
        (_HANA_TN_PRETTY_FUNCTION_TYPE_PREFIX,
         _HANA_TN_PRETTY_FUNCTION_TYPE_SUFFIX)
        * Also provides a compiler agnostic _HANA_TN__PRETTY_FUNCTION__

    * type_name_compiler_capabilities.hpp :
        Defines _HANA_TN_CAN_CONSTEXPR for MSVC and Clang
        Implements _HANA_SIZEOF_OR_STRLEN as sizeof or strlen

    * type_name_stringliteral.hpp :
        Renames `cstring` to `stringliteral`
        and adds some limited capabilities
        (_HANA_TN_MAKE_STRINGLITERAL, stringliteral_equal,
        stringliteral_equal_sz, stringliteral_to_string)

* include/boost/hana/experimental/type_name.hpp
    includes hana/experimental/detail/type_name_pretty_function.hpp
    and uses `type_name_impl_stringliteral`
    from type_name_pretty_function.hpp

* test/experimental/type_name.cpp
    contains the same tests as before, but it will be run under MSVC, Clang,
    and GCC (runtime only).
   There are two adaptations for MSVC, without a regex because it makes
   it easier to understand the kind of output MSVC produces.

* test/type_name_stringliteral_test.cpp
    contains the same tests, except that
    they are independent from Boost.Hana

    It also adds another test concerning the fact that
    __PRETTY_FUNCTION__ is west const on GCC/Clang/MSVC.
    However I think this test is probably too much.

* Those tests were re-enabled under GCC and MSVC

### DIFF
--- a/include/boost/hana/experimental/detail/type_name_compiler_capabilities.hpp
+++ b/include/boost/hana/experimental/detail/type_name_compiler_capabilities.hpp
@@ -1,0 +1,29 @@
+/*
+@file
+Defines _HANA_TN_CAN_CONSTEXPR and related macros
+(_HANA_TN_CONSTEXPR_IF_POSSIBLE, _HANA_SIZEOF_OR_STRLEN)
+
+@copyright Louis Dionne 2013-2017
+Distributed under the Boost Software License, Version 1.0.
+(See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
+ */
+
+#ifndef BOOST_HANA_EXPERIMENTAL_DETAIL_TYPE_NAME_COMPILER_CAPABILITIES_HPP
+#define BOOST_HANA_EXPERIMENTAL_DETAIL_TYPE_NAME_COMPILER_CAPABILITIES_HPP
+
+// only clang and MSVC support constexpr __PRETTY_FUNCTION__, gcc does not
+#if defined(__clang__) || defined(_MSC_VER)
+    #define _HANA_TN_CAN_CONSTEXPR
+#endif
+
+// in constexpr mode, strlen is equivalent to sizeof() - 1
+#ifdef _HANA_TN_CAN_CONSTEXPR
+    #define _HANA_TN_CONSTEXPR_IF_POSSIBLE constexpr
+    #define _HANA_SIZEOF_OR_STRLEN(var) sizeof(var) - 1
+#else
+    #include <cstring> // this include is not needed in constexpr mode, save compilation time
+    #define _HANA_TN_CONSTEXPR_IF_POSSIBLE
+    #define _HANA_SIZEOF_OR_STRLEN(var) strlen(var)
+#endif
+
+#endif // !BOOST_HANA_EXPERIMENTAL_DETAIL_TYPE_NAME_COMPILER_CAPABILITIES_HPP
--- a/include/boost/hana/experimental/detail/type_name_pretty_function.hpp
+++ b/include/boost/hana/experimental/detail/type_name_pretty_function.hpp
@@ -1,0 +1,59 @@
+/*
+@file
+Defines hana::experimental::type_name_details::type_name_impl_stringliteral().
+Also defines _HANA_TN__PRETTY_FUNCTION__ and related defines.
+
+@copyright Louis Dionne 2013-2017
+Distributed under the Boost Software License, Version 1.0.
+(See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
+ */
+
+#ifndef BOOST_HANA_EXPERIMENTAL_DETAIL_TYPE_NAME_PRETTY_FUNCTION_HPP
+#define BOOST_HANA_EXPERIMENTAL_DETAIL_TYPE_NAME_PRETTY_FUNCTION_HPP
+
+#include <boost/hana/experimental/detail/type_name_stringliteral.hpp>
+#include <boost/hana/experimental/detail/type_name_compiler_capabilities.hpp>
+#include <cstddef>
+
+#ifdef _MSC_VER
+#define _HANA_TN__PRETTY_FUNCTION__ __FUNCSIG__
+#else
+#define _HANA_TN__PRETTY_FUNCTION__ __PRETTY_FUNCTION__
+#endif
+
+
+#if defined(__clang__)
+    #define _HANA_TN_PRETTY_FUNCTION_TYPE_PREFIX "boost::hana::experimental::type_name_details::stringliteral boost::hana::experimental::type_name_details::type_name_impl_stringliteral() [T = "
+    #define _HANA_TN_PRETTY_FUNCTION_TYPE_SUFFIX "]"
+#elif defined(_MSC_VER)
+    #define _HANA_TN_PRETTY_FUNCTION_TYPE_PREFIX  "struct boost::hana::experimental::type_name_details::stringliteral __cdecl boost::hana::experimental::type_name_details::type_name_impl_stringliteral<"
+    #define _HANA_TN_PRETTY_FUNCTION_TYPE_SUFFIX ">(void)"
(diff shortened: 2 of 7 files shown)


pthom:
This might close https://github.com/boostorg/hana/issues/410

ricejasonf:
I recall having a regex in `hana/experimental/printable.hpp` that stripped the superfluous `struct` keyword. It's definitely not constexpr though. I was unaware of `__cdecl`.

It would be cool to normalize the output eventually, but maybe that is a bridge too far for this library.

It's just a nitpick, but `inline constexpr` is redundant as constexpr functions are implicitly inline, and inline variables are not available until C++17 (where constexpr doesn't imply inline :confused:)

pthom:
Hi, 

Thanks for your answer!

> It would be cool to normalize the output eventually, but maybe that is a bridge too far for this library.

Well, I am working on this! I'm doing it in a separate project. It would not be reasonable to do it in a constexpr mode (although I tried for a while). I will let you know one I have something ready, if you are interested. 

> It's just a nitpick, but inline constexpr is redundant as constexpr functions are implicitly inline, 

So, should I replace `inline constexpr` by only `contexpr`. Is there a risk that it becomes not C++17 compliant in the near future?

> and inline variables are not available until C++17 (where constexpr doesn't imply inline 😕)

I recall that I just had an issue with Visual Studio 2017 where generic lambdas need to be declared `inline` in a header file (otherwise you get a duplicate symbol at link time). Presently, only MSVC enforces this, but I guess this might become the norm. Am I right?

ricejasonf:
>I recall that I just had an issue with Visual Studio 2017 where generic lambdas need to be declared inline in a header file (otherwise you get a duplicate symbol at link time). Presently, only MSVC enforces this, but I guess this might become the norm. Am I right?

I was referring to the **function declarations** where `constexpr` implies inline. This is not the case for variables where C++17 introduces the `inline` specifier for **variables declarations**.

You may notice that this library has many not inline `constexpr` variables, but apparently it is okay as it just means that their address would be different (ie not an ODR issue).

Sorry, I don't mean to hijack your PR with that stuff. :sweat_smile:

pthom:
Well it seems that there is an issue with the CI and travis : my last commit ("ping travis" : https://github.com/boostorg/hana/pull/432/commits/37ab0dd4a45f2fdec9657a9acae0efff80e68e40) succeeds on my fork (see https://travis-ci.org/pthom/hana/builds/471039901), but was not launched on travis.

xqms:
Hey, I *think* I found a solution that works compile-time both on clang and gcc:
https://gist.github.com/xqms/6fd2a9bd776075df2b285be3321235cd

Interacting with `__PRETTY_FUNCTION__` in gcc (tested on 7.3.0) feels *very* buggy, though.

I will also add that I didn't test this code extensively yet.

ricejasonf:
Is this like https://github.com/Manu343726/ctti?

xqms:
Yes, similar technique, but seems to work also in static constexpr situations where ctti has problems (see #265 for discussion). Also, it returns a `hana::string` with characters in the type system, instead of a string_view-based object.

I may have time to clean this up into a proper PR for discussion later this week - but if someone else is interested in doing that, go ahead ;-)

pthom:
@xqms :  thanks for your input!

I do not have much time this sunday on this subject. 
I may have some time at the beginning of the week.

Some questions and suggestions:


#### Static constexpr situations
You wrote:
> but seems to work also in static constexpr situations where ctti has problems (see #265 for discussion)
and

Do you mean that you showed succesfully that ctti fails with gcc with in some situation and your code passes? If, some unit test showing this would be interesting!

#### gcc and `__PRETTY_FUNCTION__`feels buggy:
You wrote:
> Interacting with __PRETTY_FUNCTION__ in gcc (tested on 7.3.0) feels very buggy, though
Also would you be able to write some unit tests that fail under gcc and pass with other compilers? 
If so, they would be very useful.

#### Minimal repository for tests with various compilers

I have setup a minimal repository (with CI on travis and appveyor) for the tests
here: https://github.com/pthom/pretty_function_sandbox

There, I have setup CI tests with Clang linux, clang osx, gcc, and MSVC.

And you can see the build logs here (they curently fail with GCC, and pass with MSVC and Clang)
https://travis-ci.org/pthom/pretty_function_sandbox/builds/470722255?utm_source=github_status&utm_medium=notification

Suggestion: during development, your idea could be tested in this repo instead of the full 
boost hana repo (so that we get a faster CI feedback). Would you be open to this suggestion?


#### gcc  fix for __PRETTY_FUNCTION__ 
In the bug link you provided (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66639), it is written that a fix was posted in gcc on Nov, 19 2018, so that may be the latest gcc release 
(7.4, December 6, 2018) finally provides a constexpr `__PRETTY_FUNCTION__`. 
This is also an interesting thing to test.

xqms:
@pthom Thanks for the feedback! I'll try to provide test cases later on.

I think your idea of a separate sandbox is a good one - I'll open a PR there with my implementation.

> In the bug link you provided (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66639), it is written that a fix was posted in gcc on Nov, 19 2018, so that may be the latest gcc release
(7.4, December 6, 2018) finally provides a constexpr __PRETTY_FUNCTION__.

I don't think so, because the fix was committed to trunk. See here for gcc branches & releases: https://gcc.gnu.org/develop.html#timeline
To me, that means it'll end up in gcc 8. In my application, I'm stuck with 7.3 anyway, so I have high interest in a solution that works with that compiler.

pthom:
@ricejasonf : since you were mentioning your interest in normalising the output, would you mind giving me your opinion about the [CleanType](http://code-ballads.net/cleantype/) library on which I am working? 
I did not yet communicate actively about it, since I was waiting to have a presentable version. I am presently trying to gather informed  inputs about it. If you are interested in giving me your input, please answer me by email or by comment in the article (in order to try not to pollute this PR).

PR #440, opened 2019-02-18
[Hashable] hana::integral_constant with an enum value implements Hashable

Dwarfobserver:
Hi, this is my first contribution to Boost.Hana, so please tell me my mistakes :)

This small change implements Hashable for hana::integral_constant with an enumeration value :
It allows for example to use enumeration values as key for hana::map.


ricejasonf:
I think it failed because it found trailing spaces. The github site doesn't seem to reflect that though.

`$ LANG=POSIX grep '[[:blank:]]$'          "${sources[@]}"; [[ $? == 1 ]] || exit 1` is the last command ran in the log but with no error message.

Dwarfobserver:
Indeed, good catch ! I corrected them.

ricejasonf:
lgtm, but please "fixup" the commit so there is one commit.

@ldionne is currently at a standards committee meeting, so I don't know when he will look at this.

Dwarfobserver:
Sorry, I thought it was only for local commits ! Done :)

vyzaq:
Do you have any progress on that PR?

ReiserFlorian:
I would like to have this feature, too

include/boost/hana/hash.hpp
```diff
@@ -80,6 +80,16 @@ BOOST_HANA_NAMESPACE_BEGIN
             }
         };
 
+        template <typename T>
+        struct hash_integral_helper<T,
+            typename std::enable_if<std::is_enum<T>::value>::type
+        > {
+            template <typename X>
+            static constexpr auto apply(X const&) {
+                return hana::type_c<X>;
```
ricejasonf, line 89: Just taking a second look at this, is there a reason it does not normalize to `hana::type<hana::integral_constant<T, X::value>` as the others do?

PR #443, opened 2019-03-15
Add a is_valid test for non static member functions

sdebionne:
I am not sure if this is actually supported. My experiment with MSVC is that the compiler ICE  when
instantiating `auto boost::hana::type_detail::is_valid_impl<...>` but let see what the CI says.


ldionne:
Thanks for the added test coverage!

PR #444, opened 2019-04-22
Doc/fix cppreference concept links

tinko92:
Hi,
apprently the pages under https://en.cppreference.com/w/cpp/concept/* have all been moved to https://en.cppreference.com/w/cpp/named_req/* without automatic redirection and the original link produces a 404 status code. This PR fixes all of those links (if I didn't miss any).


ldionne:
Thanks a lot!

PR #461, opened 2020-02-07
Add option to disable adding tests to ctest

twam:
- Add cmake option BOOST_HANA_ENABLE_CTEST which configures
  whether unit tests are to be added to cmake's ctest. Default
  is ON to be backward-compatible.


ldionne:
What's the motivation for doing this?

twam:
Including the whole repo as a submodule into a cmake project and adding it with `add_subdirectory` without running the unit tests each time the project's unit test run. Other libraries provide similar options, e.g. https://github.com/google/googletest/blob/master/googletest/CMakeLists.txt#L18

PR #471, opened 2020-08-29
Remove std::is_pod deprecation warning spam that results from including Hana in C++20 and later.

tzlaine:


ldionne:
Thanks for the PR. Can you make it against `develop`?

include/boost/hana/traits.hpp
```diff
@@ -67,7 +67,9 @@ BOOST_HANA_NAMESPACE_BEGIN namespace traits {
     constexpr auto is_trivial = detail::hana_trait<std::is_trivial>{};
     constexpr auto is_trivially_copyable = detail::hana_trait<std::is_trivially_copyable>{};
     constexpr auto is_standard_layout = detail::hana_trait<std::is_standard_layout>{};
+#if __cplusplus < 202002L || defined(BOOST_HANA_CONFIG_ENABLE_DEPRECATED_IS_POD)
```
ldionne, line 70: I think I'd just drop it altogether. Being Boost and not the standard, I think it's more reasonable to be slightly more aggressive about these things. WDYT?

tzlaine:
Superceded.

PR #472, opened 2020-08-30
Fixes operator== with std::optional

ricejasonf:
- Prevents comparable operators from considering the tag types like tuple_tag
    - Allows Struct classes to implement their own comparable operators so it can
      be used with types like std::optional


ricejasonf:
I removed the tabs. I'm not sure about what I can do for

```
..\test\issues\github_460.cpp(26): fatal error C1001: An internal error has occurred in the compiler.
(compiler file 'msc1.cpp', line 1518)
 To work around this problem, try simplifying or changing the program near the locations listed above.
Please choose the Technical Support command on the Visual C++
```

:scream_cat:

include/boost/hana/detail/operators/comparable.hpp
```diff
@@ -24,17 +24,29 @@ BOOST_HANA_NAMESPACE_BEGIN namespace detail {
...
+
+    template <typename T>
+    using comparable_operators_sfinae = comparable_operators_sfinae_helper<
+        hana::tag_of_t<T>, std::remove_const_t<std::remove_reference_t<T>>>;
+
     namespace operators {
         template <typename X, typename Y, typename = typename std::enable_if<
-            detail::comparable_operators<typename hana::tag_of<X>::type>::value ||
-            detail::comparable_operators<typename hana::tag_of<Y>::type>::value
+            detail::comparable_operators_sfinae<X>::value ||
```
ldionne, line 41: How about:

```
namespace detail {
  template <typename T>
  struct has_idempotent_tag
    : std::is_same<hana::tag_of_t<T>, std::remove_const_t<std::remove_reference_t<T>>>
  { };
}
```

And then

```
std::enable_if_t<(!detail::has_idempotent_tag<X>::value && detail::comparable_operators<X>::value) ||  ...>
```

If we want to be lazy, we can define (or perhaps we already have somewhere) a lazy `and_` type trait. WDYT?

ldionne (changes requested):
Thanks a lot for working on that, Jason. It's a tricky issue.

I went ahead and tried removing the root cause of this issue, which is that `tag_of` is idempotent by default. As I was fixing errors and errors, it came back to me that Hana tags were *by design* an extension of the type of the object, and so for example that's why `hana::equal` is supposed to work out of the box on most user-defined types. So in addition to it being a major break, I think it would go against the spirit of the library to make `hana::tag_of` non-idempotent by default.

I think we need to search for a solution in the way we're providing comparison operators by default, as you did. I have a few comments about details, but the approach looks good to me.

include/boost/hana/equal.hpp
```diff
@@ -91,7 +91,10 @@ BOOST_HANA_NAMESPACE_BEGIN
     // Model for EqualityComparable data types
     //////////////////////////////////////////////////////////////////////////
     template <typename T, typename U>
-    struct equal_impl<T, U, when<detail::EqualityComparable<T, U>::value>> {
+    struct equal_impl<T, U, when<
+        detail::EqualityComparable<T, U>::value &&
+        !(Struct<T>::value && Struct<U>::value)
```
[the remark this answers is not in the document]
ricejasonf: I agree that my solution was a little ad hoc :adhesive_bandage:. Your suggestion means that the user could not call `hana::equal` in their `operator==`, but the test demonstrates that it is trivial to workaround.

include/boost/hana/detail/operators/comparable.hpp
ricejasonf: I'm not sure what you mean by "lazy". Do you mean short-circuiting the template instantiations?

ricejasonf: I used your metafunction, but the logic was a bit different. Neither of the operands could be a tag, or for some reason it compiled and always returned false because it was finding the default impl. (I think)

include/boost/hana/equal.hpp
```diff
@@ -191,7 +191,10 @@ BOOST_HANA_NAMESPACE_BEGIN
     }
 
     template <typename S>
-    struct equal_impl<S, S, when<hana::Struct<S>::value>> {
+    struct equal_impl<S, S, when<
+        hana::Struct<S>::value &&
+        !detail::EqualityComparable<S, S>::value
```
ldionne, line 196: Can you please add a test for that thing specifically? You could add it to `test/concept/struct/equal.cpp`.

ldionne: Actually, nevermind that. You already have one test for that in `github_460.cpp`.

ldionne:
I think I'm happy with this. Thanks a lot for the work!

ricejasonf:
What about the Visual C++ bug?

PR #473, opened 2020-09-02
Don't define traits::is_pod in C++20 and later, since std::is_pod is …

tzlaine:
…deprecated as of C++20.

This is a reformulation of a previous PR, now deleted.


ldionne:
For reference, the other PR was #471.

PR #477, opened 2020-12-30
Complete overhaul of CI

ldionne:
This should allow us to use more recent compilers, and test C++2a support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@
 
 language: cpp
 os: linux # Jobs are on Linux unless specified otherwise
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -4,7 +4,7 @@
 
 language: cpp
 os: linux # Jobs are on Linux unless specified otherwise
-dist: trusty # Jobs are on Trusty unless specified otherwise
+dist: xenial # Jobs are on Xenial unless specified otherwise
```
keryell, line 7: Why only `xenial`? You could use for example `focal` to get even more modern compiler and so on.

Issue #475, opened 2020-11-30
Support of c++20


fsmoke:
If i turning on support of c++20 on my compiler(msvc) i've got errors:

1>  d:\libraries\boost_1_72_0\boost\hana\traits.hpp(71,62): error C2039: 'is_literal_type': is not a member of 'std'

1>  d:\libraries\boost_1_72_0\boost\hana\traits.hpp(197,50): error C2039: 'result_of': is not a member of 'std'

because above traits removed in c++20...and std::is_pod is deprecated!

ricejasonf:
There is a fixed merged in for `std::is_pod`: https://github.com/boostorg/hana/pull/473/files

sbolding-LANL:
@ricejasonf, there is also an issue with `std::is_literal_type` that produces a warning at c++17 and error at c++20.  Should that just be removed for greater than c++14 to avoid the compiler warnings? I can submit a PR with that change.

ricejasonf:
A PR would be great! @ldionne is the one who would accept it.

I wonder why I am not seeing these errors. It must all be in that `traits.h` file.

sbolding-LANL:
We only started seeing it when going to gcc 11, but I was able to recreate it on Wandbox by just including `boost/hana.hpp`.  I will submit a PR.

jafj:
This PR is also required for MSVC to build with C++20 support enabled due to ```std::is_literal_type``` being removed.

ldionne:
This should be fixed now.

egorpugin:
`#if __cplusplus < 201703L`

msvc sets `__cplusplus` to `199711L`, so it won't work.

jafj:
> `#if __cplusplus < 201703L`
> 
> msvc sets `__cplusplus` to `199711L`, so it won't work.

If you pass the option `/Zc:__cplusplus`, it behaves properly. They haven't changed the default behaviour though due to legacy code always expecting `199711L`.

egorpugin:
Does hana set `/Zc:__cplusplus` by default and does it expose it to dependent targets?
If no, there will be errors surely. Either in unit tests or in user projects without that flag.

jafj:
I don't think it does - I set it myself in my root CMakeLists file.

ldionne:
I don't use MSVC, can one of you @jafj or @egorpugin please send a PR to add it?

jafj:
[...]

PR #487, opened 2021-05-11
remove std::is_literal_type that was deprecated

sbolding-LANL:
Supports #475

- A deprecation warning is thrown (tested on gcc11 with
std=c++17), so just remove the trait for > c++14.


ricejasonf:
It looks like the CI failed because dl.bintray.com support has sunsetted according to this: https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

PR #488, opened 2021-05-30
Add modern __cplusplus behaviour to MSVC target compile options

### LINKED ISSUE #475: Support of c++20

jafj:
Addition to further fix #475 without end-user CMake intervention

Adds /Zc:__cplusplus to MSVC compile options (as a `hana` target `INTERFACE` parameter) to force correct C++ standard reporting in `__cplusplus` variable.


jafj:
Close for more fine-grained solution

PR #490, opened 2021-06-04
Add a Boost-friendly subproject case to CMakeLists

pdimov:


CMakeLists.txt
```diff
@@ -1,3 +1,32 @@
...
+target_include_directories(boost_hana INTERFACE include)
+
+target_link_libraries(boost_hana
+  INTERFACE
+    Boost::config
+    Boost::core
+    Boost::fusion
+    Boost::mpl
+    Boost::tuple
+)
```
ldionne, line 24: So those are made available by means of a superproject including this file and defining those targets?

ldionne:
Thanks for the PR! I'd like to understand what would be required to make the main CMake file compatible with the rest of Boost instead - I think it would be easier to maintain in the future, and I'd like Hana to be a good citizen of Boost as much as possible.

CMakeLists.txt
pdimov: Usually, yes. Another supported workflow (or at least, a workflow I would like Boost to support) is the user manually including the required Boost dependencies with `add_subdirectory`, whether as submodules or with `FetchContent`.

All Boost libraries are expected to define a `Boost::libname` target.

(This should also work with a CMake-installed Boost - the installed config file for a library defines the `Boost::libname` imported target.)

pdimov:
The current `CMakeLists` only seems suited for root project use, so getting from here to there is not a small diff. Clean subprojects are easier to make Boost-compatible, although there are still many things that will be "wrong".

The project name must be `boost_libname` by convention; the version when inside the Boost superproject must be set to `BOOST_SUPERPROJECT_VERSION` (e.g. "1.77.0".) There must be a clean target `boost_libname` with an alias `Boost::libname` that other libraries (Boost and not) can link to. `target_include_directories` must be `include` to enable the superproject automatic install feature; or if not, the library can supply its own installation, but the package name must be `boost_libname`, the package version must match the Boost version, and the config file must declare the `Boost::libname` target and issue the required `find_dependency` calls that match `target_link_libraries`. (The destinations should also be "correct", i.e. those given by `GNUInstallDirs` to which CMake 3.14 defaults.)

As few options as possible must be declared, preferably none unless they really control how the library is built; as few lines of status output must be produced, preferably none. (For 140+ libraries, if each decides to add four options and print seven lines of output, the result is not hard to imagine.) Adding any compile flags is also not a good practice; these should be left to the master project. And of course `find_package(Boost)` is just asking for trouble. :-)

I'm working on a document to spell all this out, but it's not done yet.

ldionne:
Thanks for the explanation and sorry for sitting on this. Let's land this, and in the future, I can work on updating the main CMakeLists.txt to be Boost-compatible if I have time. Otherwise, I don't want to stall your effort on that.

pdimov:
Thanks for merging that; please let me know when it's in master, so that I can remove `hana` from `BOOST_INCOMPATIBLE_LIBRARIES`.

PR #492, opened 2021-08-09
C++20: require implicit conversion for hana::is_convertible by static_cast

ecatmur:
In C++20, static_cast can perform aggregate construction (__cpp_aggregate_paren_init). This breaks the example by allowing construction of Employee from its single direct base Person.

Change the maybe_static_cast control expression to require implicit conversion. Since this doesn't allow conversion to cv void, add a to_impl for cv void To.

### DIFF
--- a/include/boost/hana/core/to.hpp
+++ b/include/boost/hana/core/to.hpp
@@ -49,8 +49,8 @@ BOOST_HANA_NAMESPACE_BEGIN
         };
 
         template <typename To, typename From>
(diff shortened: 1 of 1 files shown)


ecatmur:
Of course, this change would disallow `explicit` converting constructors and conversion operators. An alternative would be to change the example (example/core/convert/is_convertible.cpp).

gjasny:
Without that patch Visual Studio `v17.3.0-pre.2.0` fails to compile the examples.

PR #497, opened 2021-10-29
Update compiler support

ldionne:
Testing with newer compilers should also surface some issues that were reported recently.


ldionne:
We don't use Travis CI anymore, closing. We could look into documenting current compiler support based on what we test in GH actions, but this can be done in a separate PR.

PR #527, opened 2024-03-02
This removes test project ID that will conflict with modular ID.

grafikrobot:
This removes the superfluous test project ID that will conflict with the modular project IDs for all libraries. When libraries are modular they will have the standard global ID of /boost/. Having colliding IDs will cause build errors.

#modular-boost https://github.com/users/grafikrobot/projects/1

### DIFF
--- a/test/Jamfile.v2
+++ b/test/Jamfile.v2
@@ -6,7 +6,7 @@ import config : requires ;
 import testing ;
 import regex ;
 
-project boost/hana :
+project :
     requirements
         <include>./_include
         <include>../include


ldionne (approved):

ldionne:
I am not sure whether you want this merged right away or whether you need to wait before you do it -- please merge at your discretion. Thanks!

ldionne:
Everything happens on `master` nowadays, `develop` isn't used and has gotten behind.

pdimov:
Well that's not supposed to happen, develop should at least be kept in sync, even if unused.

PR #537, opened 2025-12-13
avoid copies, when comparing maps

domin144:
There were two points, where copies were made:

1. When binding a map for use in `partial`.

This could be avoided, by using `std::ref`, but then either `partial` would have to handle `std::ref` in special way, by unwrapping the reference at the call point or the wrapped function would have to specify the accepted argument type explicitly to force cast.

Instead, partial was replaced with a `compare_at_key` helper struct capturing the `maps` by reference.

2. When returning the value from `find` wrapped into optional.

This is avoided by using `at_key` instead of `find`.

### DIFF
--- a/test/map/equal.cpp
+++ b/test/map/equal.cpp
@@ -10,6 +10,9 @@
 
 #include <laws/base.hpp>
 #include <support/minimal_product.hpp>
+
+#include <memory>
+
 namespace hana = boost::hana;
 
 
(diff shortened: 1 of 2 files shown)


domin144:
The alternative approaches can be checked at:
https://github.com/domin144/hana/tree/map_compare_no_copy_std_ref
https://github.com/domin144/hana/tree/map_compare_no_copy_modified_partial

test/map/equal.cpp
```diff
@@ -94,4 +97,9 @@ int main() {
             !=
         hana::make_map(p<1, 1>(), p<2, 2>())
     );
+
+    // Check if non-copyable objects can be compared
+    auto map_unique_1 = hana::make_map(hana::make_pair(hana::integral_constant<int, 0>{}, std::make_unique<int>(0)));
+    auto map_unique_2 = hana::make_map(hana::make_pair(hana::integral_constant<int, 0>{}, std::make_unique<int>(0)));
+    BOOST_HANA_RUNTIME_CHECK(map_unique_1 != map_unique_2);
```
ricejasonf: Maybe add `map_unique_1 == map_unique_1`. I am not seeing that `hana::equal` tries to short circuit identical types (or even address).

domin144: good point.

domin144:
UPDATE: I replaced `std::unique_ptr` with `ConstexprMoveOnly` to be more explicit in what is to be tested.

